PR #71, opened 2023-07-05
Golf Game Manager

aaylward:
part of the way somewhere...


cpp/cards/golf/golf.h
```diff
@@ -40,29 +46,64 @@ class Player {
...
  private:
   const std::deque<Card> drawPile;
   const std::deque<Card> discardPile;
   const std::vector<Player> players;
   const int whoseTurn;
   const int whoKnocked;
+  const std::string gameId;
+};
+
+typedef std::reference_wrapper<GameState> GameRef;
```
aaylward, line 80: is this right? should I be using `shared_pointer<GameState>` instead?

Issue #78, opened 2023-07-16
golf persistence


aaylward:

aaylward:
check out https://www.mongodb.com/developer/languages/rust/rust-mongodb-crud-tutorial/

aaylward:
https://github.com/grpc/grpc/blob/master/examples/cpp/metadata/greeter_server.cc#L45-L63

aaylward:
closed in #249

PR #239, opened 2024-08-04
add GameStoreInterface to enable golf persistence

aaylward:
- GameStoreInterface currently has InMemoryGameStore implementation for use in tests
 - escapist backed GameStore coming in future change
 - update to c++20
 - remove local state from golf::GameManager
 - add db name to client metadata in escapist c++ client
 - clean up game_manager.h
 - add new golf rpc proto (incomplete)

TODO:
 - add escapist backed GameStore
 - escapist server should read clientContext metadata for db name (ideally via Authentication header?)
 - InMemoryGameStore unit tests so we don't have to test via GameManager

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,8 +6,9 @@ common --enable_bzlmod=true
 common --disk_cache=~/bzlcache
 common --repository_cache=~/.bzl_repo_cache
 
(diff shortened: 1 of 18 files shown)


aaylward:
in service of #78

PR #242, opened 2024-08-04
read db-name in escapist

aaylward:
This updates escapist server to read client context metadata to allow clients to set their db name. At some point, some authn + authorization check should be added to make sure a caller is allowed to read/modify the requested db. This is the other half of https://github.com/muchq/MoonBase/pull/239/files#diff-f663c91b226b1b0ce0b75b8548117e68961ad72a0389bb89c3cd932be2cd6bd7

### DIFF
--- a/rust/escapist/src/escapist.rs
+++ b/rust/escapist/src/escapist.rs
@@ -8,10 +8,12 @@ use mongodb::bson::oid::ObjectId;
 use mongodb::bson::Document as BsonDocument;
 use mongodb::bson::{doc, Bson};
 use mongodb::error::{Error as MongoError, Result as MongoResult};
+#[cfg(not(test))]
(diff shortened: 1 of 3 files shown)


rust/escapist/src/escapist.rs
```diff
@@ -103,16 +109,37 @@ fn convert_hashmap_to_document(hash_map: HashMap<String, String>) -> BsonDocumen
...
+  let db_str_maybe = db_name.to_str();
+  if db_str_maybe.is_err() {
+    return None;
+  }
+  let db_str = db_str_maybe.unwrap();
+  let db_string = db_str.to_string();
+  if db_string.is_empty() {
+    return None;
+  }
+  Some(db_string)
```
aaylward, line 127: presumably there's a nice way to do all this

PR #443, opened 2024-12-14
downgrade stuff to allow prebuilt protoc toolchain to work

aaylward:
we can get back on the renovate train once rules_proto is updated for bazel 8

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,12 +1,31 @@
 startup --output_base=~/.bazelout
-common --incompatible_enable_proto_toolchain_resolution
+startup --digest_function=BLAKE3
+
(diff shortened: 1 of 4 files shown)


.bazelrc
```diff
@@ -1,13 +1,32 @@
 startup --output_base=~/.bazelout
-common --incompatible_enable_proto_toolchain_resolution
+startup --digest_function=BLAKE3
+
 common --color=yes
+
+# Allow rules to bring their own precompiled protoc
+common --incompatible_enable_proto_toolchain_resolution
```
aaylward, line 7: just needed until protobuf provides prebuilt protoc directly

PR #492, opened 2025-01-12
implement NewGame rpc 

aaylward:
- [x] add card proto mapper
- [x] test card proto mapper
- [x] service op
- [ ] service op tests
- [ ] client call
- [ ] client tests
- [x] bzl cleanup


aaylward:
continued in #504

PR #884, opened 2025-09-28
[posterize] adjust upload size limit + error handler

aaylward:
- also cargo fmt


rust/posterize/src/main.rs
```diff
@@ -119,6 +125,11 @@ async fn main() {
     let listen_address = format!("0.0.0.0:{}", &port);
 
     let app = Router::new()
+        .layer(TraceLayer::new_for_http())
+        .layer(RequestBodyLimitLayer::new(4096))
```
aaylward, line 129: oops -- wrong size limit here

PR #886, opened 2025-09-28
[server_pal] DRY up tower middleware stuff

aaylward:
this builder style tries to ensure routes get added before middleware layers [per the docs](https://docs.rs/axum/latest/axum/struct.Router.html#method.layer)


rust/posterize/src/main.rs
```diff
@@ -71,6 +61,9 @@ async fn blur_post(
             )
         })?;
 
+    let fifteen_seconds = time::Duration::from_secs(15);
+    tokio::time::sleep(fifteen_seconds).await;
```
aaylward, line 65: whoopsie

PR #931, opened 2026-01-05
[1d4] start chess.com public api client

aaylward:
Add Chess.com API client

  Implements a client for the Chess.com public API with support for:
  - Fetching player profiles
  - Fetching player statistics
  - Fetching games by month

  Also adds a generic HTTP client interface with JDK11 implementation to start with.


jvm/src/main/java/com/muchq/json/BUILD
```diff
@@ -5,6 +5,7 @@ java_library(
     srcs = ["JsonUtils.java"],
     visibility = ["//visibility:public"],
     deps = [
+        "@maven//:com_fasterxml_jackson_core_jackson_annotations",
```
aaylward: wat

jvm/src/main/java/com/muchq/chess_com_api/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@rules_java//java:defs.bzl", "java_binary")
+
+java_binary(
```
aaylward: s/binary/library

Issue #1271, opened 2026-07-31
futility/cache: LRUCache::get leaves a dangling iterator if push_front throws


aaylward:
Found by the review panel on #1270 while auditing portrait's allocation failure paths. Pre-existing in `domains/platform/libs/futility/cache/lru_cache.h`, untouched by that PR, and out of its scope — filing rather than folding in.

`LRUCache::get` promotes a key's recency by erasing its node from the recency list and then re-inserting at the front:

```cpp
m_list.erase(i->second.second);
m_list.push_front(key);
i->second.second = m_list.begin();
```

The map entry's stored iterator is invalidated by the `erase` and only made valid again by the third line. If `push_front` throws — it allocates a list node holding a copy of the key — the function unwinds with `i->second.second` still pointing at freed storage. A later `get` on the same key then calls `m_list.erase()` on that iterator.

Two things make this narrow rather than urgent:

- It needs the allocation to fail, so in practice it needs memory exhaustion.
- The caller has to come back for the *same key* afterwards, on a cache that survived the first failure.

But the second `get` is undefined behavior, not a second clean failure, and portrait's cache is keyed on `TraceRequest` — which holds the whole scene — so the node being allocated is not small. It is reachable by exactly the `std::bad_alloc` path #1267 was about.

The fix is the usual one for this shape: make the recency update strongly exception-safe by not destroying state you cannot restore. `std::list::splice` moves an existing node to the front without allocating and without invalidating the iterator, which removes the failure window entirely:

```cpp
m_list.splice(m_list.begin(), m_list, i->second.second);
// i->second.second stays valid and now points at the front
```

`insert`'s eviction path is worth a look at the same time, for the same class of problem.

Worth a test either way. A `std::list` with a throwing allocator, or a key type whose copy constructor throws on the Nth copy, makes the window reachable deterministically — the same "prove it with a control" shape the working agreement asks for, since a test that only asserts the first `get` fails would pass against the buggy version too.

PR #1275, opened 2026-07-31
Run the C++ suites under ASan, UBSan and TSan, and fix the first two things they found

### LINKED ISSUE #1273: CI: run the C++ suites under ASAN/UBSAN (and probably TSAN) — UB is currently invisible to CI
## Why

MoonBase configures no sanitizer anywhere — `.bazelrc`, `.github/workflows/`, and `scripts/` contain no `asan`/`ubsan`/`tsan` config at all. So a whole class of C++ defect is invisible to CI: the ones whose symptom is undefined behavior rather than a wrong answer.

This isn't hypothetical. It came out of the review panel on #1272, which found that `LRUCache::insert` mutated its map before its commit point, so a throw out of `evict()` left a map entry pointing into a freed list node. The panel demonstrated it under ASAN:

```
insert threw: hash
map=3 list=2 (invariant BROKEN)
ERROR: AddressSanitizer: heap-use-after-free ... in LRUCache<TKey,int>::get()
   #1 std::list<TKey>::splice(...)  stl_list.h:1663
```

The fix shipped, but **it could not be pinned by a test that CI runs**. The buggy and fixed orderings are externally identical — both leave the cache at `capacity+1`, both keep answering correctly — and the only difference is the use-after-free. The committed test (`AThrowWhileEvictingLeavesTheCacheUsable`) pins the recovery shape and says outright that it cannot see the corruption without a sanitizer.

That is a working-agreement problem, not just a coverage one: "a test beats an argument" doesn't work when the deliverable can't be a CI-run test.

The same gap covered the original #1271 bug. Reverting `get()` to the old erase + `push_front` is caught today only because the exception-safety tests happen to assert on copy counts; the *dangling iterator itself* would be silently fine in CI.

## Ask

1. Add `--config=asan` and `--config=ubsan` to `.bazelrc` (and consider `tsan` — see below).
2. Add a CI job running the C++ test targets under them.
3. Once it exists, add the negative test for #1272's evict-ordering defect that can't be written today — arm a throwing hash on the LRU key and assert the cache is clean afterward.

## Scoping questions for whoever picks this up

- **Which targets.** Running `//...` under ASAN is likely too slow and will light up third-party code (boost, opentelemetry-cpp, protobuf). A curated list of first-party C++ tests is probably the right start — `futility/cache`, `futility/rate_limiter`, `aura`, `portrait`, `golf_hub` — rather than everything.
- **Where it runs.** `build-and-test` already takes ~45 min on `ubuntu-latest` and is the long pole. A separate job, or a nightly/`main`-only schedule, likely beats adding to the critical path of every PR.
- **Suppressions.** Third-party frames will need an `asan_suppressions.txt` / `LSAN_OPTIONS`; opentelemetry-cpp and boost.context (fibers confuse leak detection) are the likely first offenders.
- **TSAN is arguably the higher-value one.** `LRUCache` is thread-safe by contract, `TracerService::trace` runs concurrently under a thread-pool transport, and `futility/cache`'s suite has five concurrency tests whose whole job is to catch races they can currently only detect by coincidence — `ConcurrentAccessDoesNotCorrupt` catches invariant breaks via a size assertion, not via a race detector. Worth doing even if ASAN proves too expensive.
- **Interaction with `-c opt`.** `.bazelrc` sets `build --compilation_mode=opt` and `test --compilation_mode=opt` repo-wide; sanitizer configs will want `-c dbg` or at least `-O1 -fno-omit-frame-pointer` to get usable stack traces.

## Related

- #1272 — where the gap surfaced; its PR description carries the "known coverage gap" note this issue is meant to close.
- #1271 — the original exception-safety bug, whose UB half is equally invisible today.

### LINKED ISSUE #1274: png_plusplus: errorHandler throws through libpng's C frames, leaking the png struct on every failed encode
Found by the ASAN gate added in #1273, within minutes of it existing. Pre-existing; nothing in the current suite could see it.

## What leaks

`//domains/graphics/libs/png_plusplus:png_plusplus_test` passes all 22 cases and then fails under `--config=asan` with six leaks, all from `PngPlusPlusTest.EmptyImageHandling` (three call sites × two libpng allocations):

```
Direct leak of 1232 byte(s) in 1 object(s) allocated from:
    #1 png_malloc_base            external/libpng+/pngmem.c:98
    #3 png_create_png_struct      external/libpng+/png.c:331
    #5 png_create_write_struct    external/libpng+/pngwrite.c:552
    #6 pngpp::MemoryPngWriter::MemoryPngWriter(int, int, int)
                                  png_plusplus.h:257
    #7 pngpp::imageToPng(image_core::Image<image_core::RGB_Double> const&, int)
                                  png_plusplus.h:490
    #8 PngPlusPlusTest_EmptyImageHandling_Test::TestBody()
```

## Why

`errorHandler` throws a C++ exception from inside libpng's C frames:

```cpp
static void errorHandler(png_structp png, png_const_charp msg) {
  (void)png;
  throw PngException(std::string("PNG Error: ") + msg);
}
```

The constructor is written as though that can't happen — it allocates, then guards the later failure paths with `setjmp` + `cleanup()`:

```cpp
png_ = png_create_write_struct(...);              // :257  allocates
info_ = png_create_info_struct(png_);             // :263  allocates
if (setjmp(png_jmpbuf(png_))) { cleanup(); throw ...; }
png_set_IHDR(png_, info_, width, height, ...);    // 0x0 -> libpng error
```

On a zero-dimension image `png_set_IHDR` raises a libpng error, libpng calls `errorHandler`, and the throw unwinds straight past the `setjmp` — which is never reached, because `setjmp` only returns non-zero via `longjmp`. The object was never fully constructed, so `~MemoryPngWriter` never runs and `cleanup()` never happens. Both structs leak.

The same shape exists in the file-based `PngWriter` (`errorHandler` at `:139`, `cleanup()` at `:127`) and in the readers.

## Why it's more than a leak

Throwing through libpng's C frames is undefined behavior in its own right — those frames are compiled without unwind tables, and libpng's `setjmp`/`longjmp` machinery exists precisely because it expects the error handler *not* to return normally. libpng's own internal allocations are stranded too.

This is on portrait's render path, not a corner: `TracerService::trace` catches `PngException` out of `imageToPng`, and that catch clause is load-bearing (#1267, #1271). Every render that fails inside libpng takes this path.

## Fix shape

Make `errorHandler` `longjmp` rather than throw, stash the message somewhere the `setjmp` side can read, and put every libpng call that can error under a `setjmp` — then throw the `PngException` from C++ frames where unwinding is defined. That is a rework of error handling across `PngWriter`, `MemoryPngWriter`, and the readers, with its own tests, which is why it isn't folded into #1273.

## Meanwhile

`png_plusplus_test` is tagged `no-sanitize` so the new sanitizer job can be green and useful; removing that tag is the acceptance test for this issue.

## Related

- #1273 — the sanitizer CI that found it
- #1271, #1267 — the `PngException` catch path this sits under

aaylward:
Closes #1273. Closes #1274.

MoonBase configured no sanitizer anywhere, so a whole class of C++ defect was invisible to CI — the ones whose symptom is undefined behavior rather than a wrong answer. Mutation checking can't reach them: it assumes the bug produces an observable difference, and these produce a crash somewhere unrelated, or nothing.

## The configs

`--config=asan`, `--config=ubsan`, `--config=tsan` in `.bazelrc`, none implied by anything else. All three override the repo-wide `-c opt` with `-c dbg` — a report whose stack has been inlined away names the wrong function.

Two flags worth explaining rather than just reading:

- **`-fno-sanitize-recover=all`** on UBSan. It diagnoses and keeps going by default, so a finding would scroll past inside a test that still reports green.
- **`-fno-sanitize=vptr`**, the one check needing `libclang_rt.ubsan_standalone_cxx`. The toolchain links through `clang`, not `clang++`, so leaving it on fails at link with undefined `__ubsan_vptr_type_cache` rather than finding anything. Everything that carries the weight — signed overflow, shifts, alignment, null deref, bogus bool/enum loads, float casts — is still on.

## The CI job

Its own job, not more flags on `build-and-test`: that one is already the long pole, and a sanitized build is a separate configuration with its own object tree. Its own disk-cache keys too — these artifacts are `-c dbg` and instrumented, so sharing would evict the opt ones `build-and-test` depends on.

**All three sanitizers run, as a matrix.** The configs share no build output, so sequential steps would serialize three full builds for no reuse; ASan and TSan can't be linked into the same binary regardless. `fail-fast` is off so one sanitizer's finding doesn't hide the other two, and each leg reports its own pass/fail. Disk cache is keyed per sanitizer so the legs don't evict each other.

**Scope is a build-cost line, not a compatibility one**, and I measured it both ways rather than assuming. The light set (first-party tests whose closure is also first-party) is 16 targets: ~110s under UBSan, 137s under TSan. golf_hub's race tests and portrait's suite *also* build and pass under ASan — they just add a full instrumented rebuild of boost and opentelemetry-cpp. My first attempt at the full set reported `FAILED TO BUILD`, which turned out to be **this sandbox running out of disk**, not the sanitizer. Worth chasing down before it became a false rationale in a comment.

## What it found in its first two minutes: #1274

`png_plusplus`'s error handler threw a C++ exception from inside libpng's C frames. libpng requires that handler not to return and provides `longjmp` for the purpose; throwing instead unwinds frames compiled without unwind tables and strands everything libpng held. `MemoryPngWriter`'s constructor allocates the png struct then throws before finishing, so the destructor never runs and `cleanup()` never happens.

**All 22 tests in that suite pass.** The leak report arrives after them. And it's on portrait's render path — `TracerService::trace` catches `PngException` out of `imageToPng`, and that catch is load-bearing (#1267, #1271), so every render failing inside libpng took this path.

Both writers now record the message and `longjmp`, throwing from the C++ side of the jump where unwinding is defined.

**The second half wasn't in the issue and is subtler:** `longjmp` runs no destructors. `writeImage` built its row buffers *after* the `setjmp`, so a jump out of `png_write_image` skipped those too. Buffers are now built by the caller, with the `setjmp` in a small `writeRows` helper holding nothing but a pointer — the jump lands *below* the owning frame instead of past it.

`LibpngsOwnMessageSurvivesTheJump` pins that libpng's diagnostic still reaches the caller. Nothing else would notice it going missing: the existing tests assert only that *something* throws, so a handler that recorded nothing would hand back the generic fallback and stay green.

The two *readers* are untouched. They pass `nullptr` for the error function, so they use libpng's default handler and never had this bug. They do share the destructor-skip hazard, but only on a read failure after the row buffers are allocated, and fixing it properly needs RAII around the raw `png_structp`/`FILE*` — a separate change.

## And the one that motivated #1273

`AThrowWhileEvictingLeavesTheCacheUsable` gained the assertion it couldn't carry before. `contains()` only looks in the map; `get()` follows the stored iterator into the recency list — the pointer the pre-#1272 ordering left dangling. Against that ordering:

| Build | Result |
|---|---|
| ordinary | **segfault, no gtest verdict, no diagnosis** (5 runs, 5 crashes) |
| `--config=asan` | `heap-use-after-free`, both stacks |
| `--config=ubsan` | member access within misaligned address |
| `--config=tsan` | `heap-use-after-free` |

Note the first row, because it corrects something I wrote in #1272's description: the new assertion makes the bug *detectable* without a sanitizer, just only as a mystery crash. What the sanitizers add is the **diagnosis** — plus the cases where UB doesn't happen to crash at all.

## Verification

| | |
|---|---|
| 16 first-party targets — ASan, UBSan, TSan | green on all three |
| png_plusplus, portrait, futility, aura — ordinary build | 16/16 |
| golf_hub race tests under ASan | pass (built locally, out of CI scope for cost) |
| `//:buildifier_test`, `scripts/format-all` | green |
| `branch.yml` | parses; matrix expands to asan/ubsan/tsan, `fail-fast: false` |
| Mutation testing | png fix killed 3 ways — reverting `longjmp` to a throw dies under ASan; clearing the recorded message or hardcoding the fallback both die in the ordinary build |

The job fails loudly if the target query matches nothing, so a mis-edited query reports a build error rather than a vacuous pass.

**The review panel did not run on this change.**

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -74,6 +74,56 @@ common --test_output=errors
 # don't block on remote-cache uploads
 common --bes_upload_mode=fully_async
 
+# Sanitizers (#1273). Off by default and never implied by another config:
+# they change codegen, and a sanitized binary is not what we ship.
+#
+# The point of these is the class of bug whose symptom is undefined behavior
+# rather than a wrong answer, which no assertion can see. #1272 shipped a
+# use-after-free fix whose test could only pin the recovery shape, because
+# the buggy and correct orderings are indistinguishable from outside.
+#
+# compilation_mode=dbg deliberately overrides the repo-wide opt above: a
+# report whose stack has been inlined away names the wrong function, which
+# is the difference between a lead and a shrug.
+build:sanitizer --compilation_mode=dbg
+build:sanitizer --strip=never
+build:sanitizer --copt=-g
+build:sanitizer --copt=-fno-omit-frame-pointer
+build:sanitizer --copt=-fno-optimize-sibling-calls
+# Sanitized runs are slow enough to trip the default small/medium timeouts
+# on tests that pass comfortably under opt.
+build:sanitizer --test_timeout=120,600,1800,3600
+# Targets with a known finding that has an issue and no fix yet. Each one
+# carries the issue number at its tags= line; removing the tag is that
+# issue's acceptance test. Pass --test_tag_filters= to see them anyway.
+test:sanitizer --test_tag_filters=-no-sanitize
+
+build:asan --config=sanitizer
+build:asan --copt=-fsanitize=address
+build:asan --linkopt=-fsanitize=address
+test:asan --test_env=ASAN_OPTIONS=detect_stack_use_after_return=1:strict_string_checks=1
+
+build:ubsan --config=sanitizer
+build:ubsan --copt=-fsanitize=undefined
+build:ubsan --linkopt=-fsanitize=undefined
+# Everything except the vptr check, which is the one UBSan check that needs
+# the C++-specific runtime (libclang_rt.ubsan_standalone_cxx). The toolchain
+# links through clang rather than clang++, so leaving it on fails at link
+# with undefined __ubsan_vptr_type_cache rather than finding anything. The
+# checks that carry the weight here — signed overflow, shifts, alignment,
+# null deref, bogus bool/enum loads, float casts — are all still on.
+build:ubsan --copt=-fno-sanitize=vptr
+build:ubsan --linkopt=-fno-sanitize=vptr
+# UBSan diagnoses and keeps going by default, so a finding scrolls past
+# inside a test that still reports green. Make it fatal.
+build:ubsan --copt=-fno-sanitize-recover=all
+test:ubsan --test_env=UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1
+
+build:tsan --config=sanitizer
+build:tsan --copt=-fsanitize=thread
+build:tsan --linkopt=-fsanitize=thread
+test:tsan --test_env=TSAN_OPTIONS=halt_on_error=1:second_deadlock_stack=1
+
 # Platform-specific configurations
 build:macos --features=-module_maps
 
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -168,12 +168,22 @@ jobs:
   # Beast, opentelemetry-cpp and Postgres targets do build and pass under
   # ASan; they just add a full instrumented -c dbg rebuild of boost and otel
   # to a runner that has to free disk before it can build at all. Widening
-  # this, and adding the tsan config that already exists in .bazelrc, are
-  # runner-budget calls rather than technical ones.
+  # that is a runner-budget call rather than a technical one.
+  #
+  # A matrix rather than three steps in one job, for two reasons. The three
+  # configs share no build output — each is a distinct object tree, so
+  # sequential steps would serialize three full builds for nothing — and
+  # fail-fast is off so one sanitizer's finding does not hide the other two.
+  # ASan and TSan are mutually exclusive at link time anyway.
   sanitize:
+    name: sanitize (${{ matrix.sanitizer }})
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/muchq/moonbase/ci-base:latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [asan, ubsan, tsan]
 
     steps:
       - uses: actions/checkout@v7
@@ -190,10 +200,11 @@ jobs:
       - uses: bazel-contrib/setup-bazel@0.19.0
         with:
           bazelisk-cache: true
-          # Its own disk-cache key: these are -c dbg and instrumented, so
-          # sharing build-and-test's cache would evict the opt artifacts that
-          # job depends on for every one of these it stores.
-          disk-cache: ${{ github.workflow }}-sanitize
+          # A key per sanitizer. Separate from build-and-test's because these
+          # are -c dbg and instrumented, and separate from each other's
+          # because the three configs share no object tree — one key would
+          # just have them evicting each other on every run.
+          disk-cache: ${{ github.workflow }}-sanitize-${{ matrix.sanitizer }}
           repository-cache: true
           bazelrc: |
             build --local_resources=cpu=4
@@ -215,11 +226,8 @@ jobs:
           echo "targets=$TARGETS" >> "$GITHUB_OUTPUT"
           echo "Sanitizing: $TARGETS"
 
-      - name: AddressSanitizer
-        run: bazel test --config=asan ${{ steps.targets.outputs.targets }}
-
-      - name: UndefinedBehaviorSanitizer
-        run: bazel test --config=ubsan ${{ steps.targets.outputs.targets }}
+      - name: Test under ${{ matrix.sanitizer }}
+        run: bazel test --config=${{ matrix.sanitizer }} ${{ steps.targets.outputs.targets }}
 
   # Metal GPU training requires macOS; run microgpt tests on Apple Silicon when relevant files change.
   test-microgpt-metal:
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -153,6 +153,74 @@ jobs:
         run: |
           ./scripts/diff-build origin/main
 
+  # Sanitizers (#1273). The class of bug these catch — use-after-free, leaks,
+  # signed overflow — has no symptom an assertion can see, so before this job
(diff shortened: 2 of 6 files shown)


aaylward:
`build-and-test` is red on `//domains/games/apis/golf_hub:pg_hub_e2e_test` — **TIMEOUT in 60.0s**, 135/136 otherwise green. Not caused by this PR, and not something I'm going to paper over. Filed as #1276.

Short version: this is the recurrence that the test's `size = "small"` ceiling was deliberately kept tight to catch, second sighting after #1241. The ceiling did its job.

Five tests in that fixture pass in ~276ms, then `TwoInstancesShareOneGame` — the cross-instance LISTEN/NOTIFY test from #1224 — hangs for the full 59s after both hubs report `hub restored 0 rooms, 0 members, 0 games`.

The diagnostic bit is the Postgres service log: **nothing at all between 20:35:49 and the 20:39:19 checkpoint**, and the test ran 20:37:17→20:38:16. The database was idle for the entire hang. No slow query, no lock wait, no connection churn. So it isn't runner I/O or a slow checkpoint — it's a lost wakeup, with the test blocked on a cross-instance event that never arrived. #1276 has the full evidence and the suspected shapes.

Two things about this PR specifically:

- **Nothing in the diff touches golf_hub or pg.** The six changed files are `.bazelrc`, `branch.yml`, `docs/TESTING.md`, `png_plusplus.{h,_test.cc}` and `lru_cache_test.cc`. I checked the `.bazelrc` diff for leakage into the default build: every line is namespaced under `build:sanitizer` / `build:asan` / `build:ubsan` / `build:tsan`, nothing at `common` or bare `build` scope.
- **It is why the test ran at all.** Editing `.bazelrc` invalidates the graph, so `diff-build` ran the full 136-test suite instead of a subset. On a typical PR these pg-backed golf_hub tests don't execute — which is also why sighting #2 took this long to arrive.

All three new sanitizer legs passed on their first CI run: `sanitize (asan)` 10m05s, `sanitize (ubsan)` 9m49s, `sanitize (tsan)` 11m20s. Every other job is green.

I don't have `actions: write` on this token, so I can't re-run the failed job myself — happy to push a no-op commit to re-trigger if you'd rather not click it.

---

Issue #1282, opened 2026-08-01
one_d4: bound how long a single indexing run may hold its lease


aaylward:
Follow-up from #1279. Not a regression it introduced — the gap is a consequence of the shape reclamation now has, and it is the one case the liveness probes cannot see.

## The gap

`IndexingRequestDao.reclaimStale` settles a request three ways: released (lease lapsed, attempts remain), poisoned (claimed `MAX_ATTEMPTS` times), stalled (unheld, untouched for `STALE_REQUEST`, and no worker anywhere holds a live lease).

Every one of those requires the owner to have *stopped*. A worker that is wedged mid-run has not stopped. `IndexWorker` renews from a separate heartbeat thread, so a run blocked on a hung socket read, a deadlocked pool, or an unbounded retry keeps its lease live indefinitely:

- the row is held with an unexpired lease, so the released arm skips it
- `attempts` never advances, because nothing re-claims it, so the poisoned arm skips it
- the fleet-liveness probes in `RETIRE_ABANDONED_SQL` both see that live lease, so the stalled arm skips it — and, worse, that same live lease suppresses the stalled arm for *every other* queued row in the table

So one wedged run can hold both its own range and the whole backlog's answer, with no bound on how long.

This is narrower than it was before #1279 in one way and wider in another. Narrower: a wedged process used to take its process-local queue with it, so only that instance's work was affected. Wider: the fleet-liveness probe means one wedged worker anywhere now suppresses the backstop for the entire table.

## Why a better probe does not fix it

The probes answer "is a worker alive?" and the worker genuinely is. The unanswerable question is "is this run making progress?", which needs either a ceiling on run duration or progress reporting, not a liveness signal. `RETIRE_ABANDONED_SQL`'s javadoc already says as much and defers to this issue.

## Sketch

Options, roughly in order of how much they cost:

1. **Cap total lease renewals per claim.** The heartbeat stops renewing after `MAX_RUN_DURATION`; the lease then lapses and the normal released → poisoned path takes over. Cheapest, entirely inside `IndexWorker`, and it reuses machinery that already exists. Needs the cap to be generous — a twelve-month range over a slow chess.com is legitimately long.
2. **Interrupt the run at the cap**, rather than only ceasing to renew. Stops the wedged work actually consuming a worker slot, but needs the extraction executor and the HTTP client to be interruptible, which wants checking.
[...]

Issue #1250, opened 2026-07-29
one_d4: stranded PENDING indexing_requests poison dedupe forever


aaylward:
`IndexingRequestDao.findExistingRequest` matches rows with `status IN ('PENDING','PROCESSING')` with **no age cutoff**, and nothing ever reconciles rows whose owner died. A stranded row permanently satisfies dedupe: every subsequent identical submit — REST and MCP alike — returns the dead row's PENDING status instead of doing work, until someone passes `skipCache: true`.

## Ways a row gets stranded

1. **Process restart with queued work.** `InMemoryIndexQueue` is process-local and volatile while `indexing_requests` is durable Postgres. Any restart with messages still queued orphans their PENDING rows. Now that mcpserver embeds the indexer as a second process class writing to the shared table, restarts are twice as likely to leave orphans.
2. **Inline processing failure (MCP hybrid path).** `IndexRequestService.start` creates the row, then calls `inlineProcessor.accept(message)` without enqueueing. `IndexWorker.process` catches `Exception` and marks FAILED, but if the initial `updateStatus(..., "PROCESSING", ...)` throws (transient DB blip), the catch-block `updateStatus(..., "FAILED", ...)` throws, or an `Error` escapes, the exception propagates to the MCP caller and the row stays PENDING with no queue entry and no retry owner. (Current behavior is pinned by `IndexRequestServiceTest.submitHybrid_inlineProcessorThrowingPropagatesAndLeavesRowPending`.)

## Suggested fixes (any subset)

- Age-bound the dedupe match, e.g. `AND updated_at > now() - interval '1 hour'` (or an app-side cutoff) so dead rows stop matching.
- On startup, sweep PENDING rows with no in-memory queue entry: re-enqueue or mark FAILED with "orphaned by restart".
- In `IndexRequestService`, wrap the inline dispatch so an escaping exception best-effort marks the row FAILED before rethrowing.

Related: #1249 covers the create-side race; this one is about rows that already exist but will never complete.

Found by review panel on #1248.

PR #1283, opened 2026-08-01
Let any worker run any queued indexing request (#1279)

### LINKED ISSUE #1279: one_d4: any available worker should be able to process any queued request (dispatch from the table, not an in-memory queue)
Today a request can only ever run in the process that accepted the submit. `IndexRequestService` calls `createOrAdopt`, then either hands the message to `InMemoryIndexQueue` — drained by a single thread in that same JVM — or runs it inline. Nothing else can pick it up.

That is the binding constraint on scale-out, and it has three consequences now:

- **Adding an instance adds no throughput for work already queued.** The queue is process-local, so a busy instance stays busy while an idle one has nothing to take.
- **The documented two-JVM deployment is already uneven.** REST and MCP are separate processes against one Postgres whenever `INDEXER_DB_URL` is set. A request submitted to MCP runs on MCP, however loaded it is.
- **A restart loses queued work outright.** The rows survive, the messages do not. They sit PENDING with nobody to run them until reclamation retires them, and the user is told to re-submit.

There is also a correctness gap that only closes with this change, noted in `RetentionPolicy.STALE_REQUEST` and `IndexingRequestStore.reclaimStale`: a queued message has no owner renewing for it, so a backlog deeper than the one-hour cutoff retires work that is owned and about to run. Raising the cutoff trades one wrong answer for a slower one. The fix is for there to be no such thing as work that is queued but invisible.

### Shape

`indexing_requests` becomes the queue. #1278 already added the hard part — `owner_id`, `lease_expires_at`, and a `claim` that is a single conditional UPDATE, so two instances racing for one row cannot both win. What is missing is the *selection* step:

```
claimNext(ownerId, lease, now) -> Optional<IndexingRequest>
```

Select live rows that are unclaimed or whose lease has expired, oldest first, then attempt `claim(id, ...)` on each candidate until one succeeds. The existing conditional UPDATE is what makes that safe without `FOR UPDATE SKIP LOCKED`, which H2 does not support — worth keeping the portable shape rather than splitting the implementation by dialect.

Workers then run a poll loop. The in-memory queue should stay, but demoted to a local wake-up nudge so the inline/hybrid latency path is unchanged: losing it costs latency, not work.

### Three things this needs decided, not just implemented

1. **`skipCache` is not persisted.** `IndexMessage` carries player, platform, months, `excludeBullet` and `skipCache`. Every field but the last is a column on `indexing_requests`. A worker that picks a row up off the table cannot know whether the submitter asked to bypass the period cache, so `skip_cache` has to become a column — otherwise dispatch silently changes behaviour for those requests.

2. **Poison pills.** Once any worker can take any request and reclamation hands an expired lease to the next one, a request that reliably kills its worker is retried forever, across the fleet. Needs a bounded attempt count on the row, and a terminal state when it is exhausted. This does not exist today because a crashed process took its queue with it.

3. **What `STALE_REQUEST` means afterwards.** The one-hour arm currently means "nobody ever picked this up, so it is orphaned". With table-backed dispatch an unclaimed row is *queued*, not orphaned, and retiring it is wrong. The arm either goes away or narrows to something that is genuinely unrunnable. Whichever, the two-clock split from #1278 and the README retention table need updating with it.

### Verification

The property to pin is the one in the title: a request submitted to instance A is processed by instance B when A is not running it. That is testable against H2 with two `IndexWorker` instances and one store, and the poison-pill bound and the `skip_cache` round-trip are both ordinary unit tests.

Depends on #1278.

aaylward:
Closes #1279.

`indexing_requests` becomes the work queue. `IndexWorkerLifecycle` claims the oldest row nobody holds, takes a lease, and runs it — whichever instance accepted the submit.

Before this, a request could only be processed by the JVM that accepted it, because the message went into that process's `InMemoryIndexQueue`. Adding an instance added no throughput for work already queued; a restart lost the messages while the rows survived; and the documented two-JVM deployment (one_d4 REST plus mcpserver against one Postgres) sent load wherever the submit happened to land. The queue survives, demoted to a wake-up nudge for the instance that took the submit — losing one costs latency, not work.

Claiming is what makes concurrent pollers safe, and #1280 already built it: every worker has an owner id and every write is fenced on it. Selection then conditional UPDATE rather than one statement, because H2 has no `SELECT ... FOR UPDATE SKIP LOCKED` and the conditional UPDATE is what makes it safe without one — at most one racer's WHERE can match, and losing costs a retry against the next candidate. Raced on real Postgres 16, not only H2.

## Reclamation had to be rethought, not extended

An expired lease used to mean "retire it": nothing was ever going to run that request again. Now the row *is* the queue, and retiring throws away work any worker could pick up. Three arms, applied **poisoned → stalled → released**:

| Situation | Outcome | Why |
|---|---|---|
| Lease lapsed, attempts remain | **Released** | The owner is gone; the work is not. Clear `owner_id`, keep the row and its `dedupe_key`. Silent — the user is not told about work that is about to run. |
| Claimed `MAX_ATTEMPTS` times, each worker stopping short | **FAILED** (poisoned) | Releasing is unbounded by construction, so without a counter a request that kills its worker tours the fleet forever, costing one per lap — impossible while a crashed process took its queue down with it. |
| Unheld, untouched for `STALE_REQUEST`, and no worker anywhere holds a live lease | **FAILED** (stalled) | Every instance is down or partitioned. The user is owed an answer and silence is the worse one. |

Both bounds are needed and neither substitutes for the other. Attempts alone never answers a user whose whole fleet is dead. The age window alone cannot tell "nothing is serving this" from "my turn hasn't come" — one worker draining a deep backlog leaves rows at the back untouched for as long as the backlog takes.

**The fleet-liveness probe took three tries.** Sampling "any live lease right now" is not enough: leases are held in bursts, and the real gap between one job's terminal write and the next claim FAILED forty-nine queued rows in one statement — worse than the bug the guard exists to prevent. "Anything touched lately" is worse still, because it counts submits, so a user submitting into a dead fleet suppresses their own answer indefinitely. The shipped form requires **both** probes, reads `lease_expires_at` because only a worker sets it, and excludes the row being judged. Terminal writes and releases now leave that column alone so it survives as an honest record of when a worker last held the row.

Arm order is pinned by a test, and for the two reasons that are actually load-bearing — not the one that looks obvious. A spent row cannot match the release arm at all (that statement carries its own attempts guard), so this is not about stopping an extra lap. Poisoned before stalled because both retire a spent row but only one tells the user why. Stalled before released because releasing stamps `updated_at`, which would make the row look freshly touched and hide it from the staleness check for another whole window.

## The submit path now asks one question

`createOrAdopt` and nothing else. A `findExistingRequest` read used to short-circuit ahead of it as an optimization; once that read stopped judging staleness, a row nobody would ever run still read as live, so the short-circuit answered every resubmit with it and never reached the reclaim that retires it. Both paths converged on `createOrAdopt` anyway — the read only ever saved the settle, which is precisely what it must not save. An existing #1250 regression test is what caught this.

That leaves `findExistingRequest` with no production caller, documented under the same rule `holdsLease` is: reading it and acting on the answer is the check-then-act `createOrAdopt` exists to close.

## Also here

- A worker no longer runs the same request twice in one process — the nudge and the poll could both reach it.
- Nudges are drained on the working path. A busy instance never reaches the blocking poll, so it would otherwise bank stale messages and fire them back-to-back the moment things went quiet.
- Schema: `skip_cache` and `attempts` on `indexing_requests`, both backfilled by their column DEFAULTs, plus a claimable-rows index — partial on Postgres, composite on H2, which has no partial indexes.
- `RetentionWorker` logged "retired N stranded requests" for a count that is mostly requeues now.
- `IndexRequestServiceTest`'s fake retired on lease expiry — the exact behaviour this change deletes — so a service that resurrected released work would have looked correct. It now models all three arms including the liveness probes, and its `claimNext` throws rather than returning empty, so a later change that makes the submit path claim fails loudly instead of reading as "no work".

## Verification

- `bazel test //domains/games/apis/one_d4/...` — 52/52.
- Same suite against a **real Postgres 16** with `PG_TEST_DB_URL` set — 52/52. The PG-gated suites skip silently without it, so a green local run proves nothing about the SQL; several defects here were Postgres-only.
- `scripts/diff-build origin/main` — 40 impacted targets, all passing, including the downstream `mcpserver`.
- Arm ordering, attempt exhaustion, and the `flushOwned` row lock are each pinned by a test that fails when the behaviour is reverted; found via `scripts/mutation-check` after two earlier versions of the concurrency tests turned out to be vacuous.
- Review panel ran on the implementation and produced two of the fixes above (the same-process double-run and the liveness probe). It did **not** run on this final cleanup pass; the one behavioural change in it — removing the dedupe pre-read — was caught and is pinned by an existing regression test.

## Known gap

Filed as #1282: a worker wedged mid-run keeps renewing from its heartbeat thread, so both probes stay positive and that one lease suppresses the stalled arm for the whole table, not just its own row. No liveness probe can close it — the worker is alive; what is unknown is whether the run is progressing — so it needs a ceiling on run duration. Referenced from `RETIRE_ABANDONED_SQL`.

---
_Generated by [Claude Code](https://claude.ai/code/session_012URbCariWNbKb58XsrQtSA)_

### DIFF
--- a/domains/games/apis/mcpserver/BUILD.bazel
+++ b/domains/games/apis/mcpserver/BUILD.bazel
@@ -174,6 +174,7 @@ java_test_suite(
         ":dtos",
         ":module",
         ":tools",
+        "//domains/games/apis/one_d4:worker",
         artifact("io.micronaut:micronaut-context"),
         artifact("io.micronaut:micronaut-inject"),
         artifact("org.junit.jupiter:junit-jupiter-api"),
--- a/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/McpModule.java
+++ b/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/McpModule.java
@@ -179,7 +179,11 @@ public class McpModule {
         indexExtractionExecutor);
   }
 
+  // preDestroy is load-bearing, not hygiene: the poller is a daemon thread, so without it a deploy
+  // is indistinguishable from a crash — the row stays owned by a dead process for a full lease, and
+  // the attempt it spent is gone. See IndexWorkerLifecycle#stop.
   @Context
+  @Bean(preDestroy = "stop")
   public IndexWorkerLifecycle indexWorkerLifecycle(
       IndexQueue queue, IndexWorker worker, IndexingRequestStore requestStore, Clock clock) {
     return new IndexWorkerLifecycle(queue, worker, requestStore, clock);
--- a/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/McpModule.java
+++ b/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/McpModule.java
@@ -180,8 +180,9 @@ public class McpModule {
   }
 
   @Context
-  public IndexWorkerLifecycle indexWorkerLifecycle(IndexQueue queue, IndexWorker worker) {
-    return new IndexWorkerLifecycle(queue, worker);
+  public IndexWorkerLifecycle indexWorkerLifecycle(
+      IndexQueue queue, IndexWorker worker, IndexingRequestStore requestStore, Clock clock) {
+    return new IndexWorkerLifecycle(queue, worker, requestStore, clock);
   }
 
   @Context
--- a/domains/games/apis/mcpserver/src/test/java/com/muchq/games/mcpserver/McpModuleTest.java
+++ b/domains/games/apis/mcpserver/src/test/java/com/muchq/games/mcpserver/McpModuleTest.java
@@ -4,6 +4,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import com.muchq.games.mcpserver.dtos.Tool;
 import com.muchq.games.mcpserver.tools.ToolRegistry;
+import com.muchq.games.one_d4.worker.IndexWorkerLifecycle;
 import io.micronaut.context.ApplicationContext;
 import org.junit.jupiter.api.Test;
 
@@ -13,6 +14,29 @@ import org.junit.jupiter.api.Test;
  * only by manual smoke tests.
  */
 public class McpModuleTest {
+
+  /**
+   * The container's half of the shutdown contract, exercised for real: boot, then close, and the
+   * poller must have been told to stop.
+   *
+   * <p>Worth a real context rather than reading the annotation back off the factory method. The
+   * poller is a daemon thread, so if Micronaut does not invoke preDestroy — wrong bean scope, a
+   * misspelled method name, a refactor that renames stop() — nothing anywhere fails, and every
+   * deploy silently strands its in-flight request for a full lease and spends one of its three
+   * attempts. This is the one place that catches that.
+   */
+  @Test
+  public void closingTheContextStopsTheIndexWorker() {
+    IndexWorkerLifecycle lifecycle;
+    try (ApplicationContext context = ApplicationContext.run()) {
+      lifecycle = context.getBean(IndexWorkerLifecycle.class);
+      assertThat(lifecycle.isRunning()).as("the poller should be live while the app is").isTrue();
+    }
+
+    assertThat(lifecycle.isRunning())
+        .as("shutdown must reach the poller, or every deploy looks like a crash")
+        .isFalse();
+  }
 
   @Test
   public void contextStartsAndRegistersAllTools() {
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -378,6 +378,7 @@ java_test_suite(
     name = "worker_tests",
     size = "small",
     srcs = [
+        "src/test/java/com/muchq/games/one_d4/worker/IndexWorkerLifecycleTest.java",
         "src/test/java/com/muchq/games/one_d4/worker/IndexWorkerTest.java",
         "src/test/java/com/muchq/games/one_d4/worker/OpeningsTest.java",
         "src/test/java/com/muchq/games/one_d4/worker/RequestLivenessTest.java",
@@ -474,6 +475,7 @@ java_test_suite(
         "src/test/java/com/muchq/games/one_d4/db/IndexedPeriodDaoTest.java",
         "src/test/java/com/muchq/games/one_d4/db/IndexingRequestDaoTest.java",
         "src/test/java/com/muchq/games/one_d4/db/MigrationTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/TableDispatchTest.java",
     ],
     deps = [
         ":db",
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -460,7 +460,11 @@ java_test_suite(
         "src/test/java/com/muchq/games/one_d4/IndexerModuleTest.java",
     ],
     deps = [
+        ":db",
         ":module",
+        ":queue",
+        ":worker",
+        artifact("io.micronaut:micronaut-inject"),
         artifact("org.junit.jupiter:junit-jupiter-api"),
         artifact("org.assertj:assertj-core"),
     ],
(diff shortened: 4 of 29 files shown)


aaylward:
All three addressed in `3c221db`, test-first in each case. Thanks — finding 2 was a real overclaim in the PR body, and I've corrected it below.

## 1. Deploys took the crash path

Confirmed: `@Context` on both `indexWorkerLifecycle` beans, no `preDestroy`, so `stop()` was dead code.

Both modules now declare `@Bean(preDestroy = "stop")`, and `stop()` hands back every request this process holds through a new fenced `handBack` — clears `owner_id`, keeps `lease_expires_at`, and **returns the attempt**.

On the drain question: it hands back rather than joining, deliberately. An indexing run has no bound — a twelve-month range against a slow chess.com is legitimately long — so a shutdown that waits for one is a shutdown that hangs and gets killed anyway. Handing back delivers what a drain was for: the work moves to a surviving instance immediately, unspent, instead of after the lease. The departing run isn't interrupted and doesn't need to be — its next write is fenced on ownership it no longer has, so it unwinds itself. That's the lapsed-lease mechanism, reached a few minutes earlier.

Returning the attempt is the part that fixes what you identified. A hand-back is the evidence that separates a deploy from a crash: the worker survived long enough to say it was leaving, which is exactly what a killer request prevents.

Tests: `stopHandsBackTheRequestThisProcessWasRunning` (blocks inside `fetchGames`, calls `stop()` mid-run, asserts attempts back to zero and another owner can claim immediately), `stopIsHarmlessWhenNothingIsRunning`, plus three `handBack` tests at the store level covering the attempt, the fencing, and the lease mark surviving.

For the wiring itself, `McpModuleTest` boots a real context and closes it, so the container's half is exercised rather than assumed — `isRunning()` must be false after close. Removing the annotation fails it.

## 2. Arm ordering — you're right, and the PR body was wrong

Only stalled-before-released was pinned. Nothing constructed a poisoned+stalled overlap, and the reason is worth writing down: spending the attempts leaves `updated_at` at the last claim, which is recent by definition, so a poisoned row isn't normally stalled too. It takes `exhaustAttempts` *and* a backdate.

Added `reclaimStale_prefersThePoisonedReasonWhenARowIsAlsoStalled`. Verified by actually swapping the two `execute()` calls — it fails, and the only thing that changes is the sentence the user reads, from "this range fails repeatedly" to "indexing is down", which sends them to resubmit into a system that will fail them the same way.

## 3. `lease_expires_at` on the unowned terminal path

Fixed — that path no longer clears it, matching `updateStatusOwned`.

Narrower than the owned path but not narrow in effect, and the framing that was wrong in the code is worth naming: "this caller holds no token" is not "no worker ever claimed this". The inline handler's request was claimed and leased by the worker that then threw. Erasing its mark makes the fleet look dead, so it retires the entire queued backlog behind it in one statement — the exact failure the probe exists to prevent, through a different door.

Test is at that level rather than on the column: `reclaimStale_stillSeesAWorkerAfterAnUnownedTerminalWrite` queues three old rows, has a worker claim-and-fail through the unowned path, and asserts the sweep touches nothing. It failed with 3 rows retired before the fix.

## Nits

All four: `DESIGN.md`'s diagram redrawn (it still showed `IndexQueue → IndexWorker` as the dispatch path; now `indexing_requests → claimNext → IndexWorker` with the nudge as a dashed side-channel); `IN_PROCESS_MODE.md`'s "via the queue" and its in-memory-queue framing corrected; the `ADD_DISPATCH_COLUMNS` comment was truncated mid-sentence and is now a proper doc explaining why the DEFAULT *is* the backfill; and `upsertPeriod`'s second call site now carries the same note the 404 path had — it's unfenced, and what keeps it safe is that the `flushBatch` above it checks ownership and throws.

## Verification

Every new assertion mutation-checked, all killed:

- unfencing `handBack` → `handBack_refusesOnceAnotherWorkerHasTakenTheRequest`
- `attempts = attempts` instead of the decrement → `handBack_returnsTheRequestWithoutSpendingAnAttempt`
- clearing `lease_expires_at` on either terminal path or on hand-back → the corresponding fleet-liveness test
- emptying the shutdown loop → `stopHandsBackTheRequestThisProcessWasRunning`
- removing `preDestroy` from either module → the respective module test

60/60 on `one_d4` + `mcpserver`, including against a real Postgres 16 with `--nocache_test_results` so the PG-gated suites genuinely ran rather than silently skipping. `scripts/diff-build origin/main` green across 40 impacted targets.

One note on process: the review panel ran on the original implementation, not on this follow-up batch — same as stated in the PR description.

---

Issue #1212, opened 2026-07-24
Observability story for Java services (one_d4, mcpserver)


aaylward:
## The gap

C++ (`aura`/`futility`) and Rust (`server_pal`) emit the shared `http_server_*` family and are on the dashboard. The Java services are dark: **one_d4** (Micronaut) and **mcpserver** pull in no micrometer/otel deps, set no `OTEL_RESOURCE_ATTRIBUTES` in compose, and aren't in prom_proxy's registry (#1199). They're deployed and invisible.

#1199 flagged "one_d4's Micronaut instrumentation is its own piece of work," and #1029 (one_d4 Production Hardening) lists an OTEL + Micrometer `/metrics` task — but scoped to one_d4 alone, as a hardening to-do, not as a shared approach. This issue is the **cross-service Java rails**: decide once how a Java service joins observability, then apply it to both (and whatever Java service comes next). #1029's one_d4 task becomes the first concrete instance of the pattern decided here.

## Goal

Bring Java onto the same standard rails as the other stacks, so no per-language special-casing downstream:

1. **Standard serving metrics.** Emit the `http_server_*` family with the *same* names, the `service_name` label, and the **microseconds** duration histogram (parity with C++/Rust — the alignment #1201 had to fix), via Micronaut/Micrometer → OTLP → the existing collector. Then #1199's standard block reads Java with zero prom_proxy changes.
2. **OTEL resource.** Set `service.name` and `service.version` (via the deploy-time mechanism in #1208 §4 — just needs `OTEL_RESOURCE_ATTRIBUTES` wired into the Java services in compose, same as the others). This also gets them the running-version + crash-loop visibility for free.
3. **Registry / catalog.** Add them to prom_proxy's `serviceOrder` so they surface as service pages.
4. **Cross-cutting families.** Once #1209 lands, emit the standard `rate_limit_*` / `cache_*` from the Java side too (one_d4 already needs a Chess.com-API rate limiter per #1029).

## Approach (decide in impl)

- `micronaut-micrometer` + `micrometer-registry-otlp`, or Micronaut's OTEL integration, exporting OTLP to `otelcol:4318`.
- Micrometer's built-in `http.server.requests` uses different names/units than our family — either remap it to `http_server_*`/microseconds or emit our instruments directly. Naming/label/unit parity is the whole point (a future conformance suite should pin this — separate issue).
- mcpserver likely wants a lighter touch than one_d4.

## Related

[...]

PR #1285, opened 2026-08-02
Java services get custom metrics; 1d4 reports indexing and motifs; Host page gets its Disk panels back (#1212)

aaylward:
Eight commits. Separable if you'd rather split them.

> **Two review rounds, a correction, and a gap closed.** A three-lens panel (correctness / data+cost / tests+claims) ran against commits 1–3; commit 4 is its findings. cursor[bot] then reviewed the result and found a blocker in commit 3 that would have taken the collector down; commits 5–6 are that round. Commit 7 corrects a claim the earlier rounds got wrong. Commit 8 closes the one thing I had declined to do. All of it is described below rather than quietly fixed.

---

# 1 — the games table crushes its own columns

Eleven columns do not fit a laptop, and the table asked for `width: 100%` with no floor. The browser honours that by squeezing the trailing columns rather than letting `.table-wrap` scroll, and the squeeze is the bug: ISO dates break after the month, and motif badges split mid-pill with `discovered attack` hanging outside its cell.

A `64rem` floor so overflow becomes a scroll; `white-space: nowrap` on cells, since every value in this table is an atom (the motifs cell opts back in — its badges are *meant* to wrap, just never mid-badge); `nowrap` on the badge; and a styled scrollbar with a stable gutter, because overlay scrollbars stay invisible until something scrolls them and an overflowing table then reads as truncated. All scoped to this table so the shorter index-requests table still fits a phone.

Verified in Chromium at 600–1700px: with the rules disabled, badges split at every width at or below 1100; with them, no badge breaks and none escapes its cell.

**Two things I could not verify.** I never reproduced clipping at ~1700px, the width the original screenshot implies — at that width the table fits with room to spare. And headless reserves no scrollbar height, so the affordance rests on the CSS rather than a measurement. jsdom has no layout engine, so the tests assert the class *and* — since commit 5 — the rules behind it, read off the stylesheet. Neither is a layout assertion.

---

# 2 — Java services could not report anything but requests

**Not a missing registry entry.** `yodel` was HTTP-only end to end: `HttpServerMetrics` exposed only `recordRequestStart/Complete/Abandoned`, and `OtlpJsonEncoder` was hard-wired to `MethodSnapshot`. There was nowhere to record "games indexed" *to* — hence `"one_d4": {}` next to a comment reading *"the Java services (#1212): yodel's standard instruments only, no custom set yet"*. `mcpserver` is empty for the same reason.

**yodel** gains `CustomMetrics` (labelled counters and distributions), an OTLP scope beside the `http_server_*` family, and `YodelMetricsFactory` so the pipeline is a shared bean application code can inject. One pipeline per process — two would post two partial views of the same service and the dashboard would double-count.

**one_d4** emits run outcomes (`completed` / `failed` / `interrupted` / `lease_lost`), games indexed, months by result (`indexed` / `degraded` / `empty` / `cached`), archive fetches (`ok` / `no_archive` / `error`), motif occurrences by motif, and distributions for run duration (labelled by outcome) and games per month. `interrupted` is the #1282 ceiling made visible.

**prom_proxy** gets Indexing and Motifs groups, rendered generically. Queries are scoped by `service_name`.

Labels are bounded vocabularies; a test sweeps every emitted series for player names and URLs.

---

# 3 — the Host page's Disk panels

Empty because **the collector was reading the wrong machine.** `otelcol` mounted only its config file — no `/proc`, no `/sys`, no host root, no `root_path` — so `hostmetrics` scraped its own container. `cadvisor`, twelve lines below, has had `/:/rootfs:ro` since it was added.

**Three of the six scrapers kept working**, which is why this survived: `cpu`, `memory` and `network` read files that exist in any namespace, so they reported plausible numbers — the container's. Only `disk` and `filesystem` produced nothing, and an empty chart looks like an idle machine.

Fixed with the documented pairing: `/:/hostfs:ro` plus `root_path: /hostfs`. **CPU and memory will now describe the host**, which is what a page called Host Metrics should have shown. Host root brings the runtime's overlay mounts along, so those and the virtual filesystems are excluded.

The mount and the `root_path` are in different files and **either alone is the broken state**, so `deploy_config_test` pins them as a pair.

---

# 4 — the panel round

## Correctness

- **Double close, introduced by this PR.** Making the pipeline a shared bean inverted its ownership, but the filter kept its `@PreDestroy`. `OtlpHttpMetricsExporter.close()` runs a final synchronous export, so shutdown fired two OTLP POSTs, each able to block for the request timeout — and Micronaut destroys the filter *before* the pipeline it was stopping.
- **A snapshot handed out the live hash key.** Mutating it stranded the entry, so the next record for those labels minted a second one — two data points with identical attributes in a payload Prometheus rejects.
- `record()` now refuses NaN, infinity and negatives, matching `add()`. A negative makes a cumulative `_sum` fall, which reads as a counter reset; protojson accepts `"NaN"` for a double, so that one parses cleanly and poisons the series forever.
- `service_name` rejected as a caller label; the `Series` comparator no longer ties on `toString`.

## Data and cost

- **`run_failure_rate` was `total - completed`.** `sum()` over no series is an *empty vector*, and vector-minus-empty is empty — so the chart vanished on a fresh process where every run was failing, which is when someone looks at it. Now a selector.
- **The motifs-per-game tile is gone.** Its two counters sit on opposite sides of the durability boundary, so an interrupted month contributes motifs and zero games, and any clamp large enough to avoid dividing by zero turns the result into a four-order-of-magnitude spike. A tile that is wrong exactly when the system is unhealthy is worse than no tile.
- **Run duration used the HTTP latency bounds** (top bound 10ms), so every observation landed in overflow: fifteen series pinned at zero and a p95 that reads a flat 10ms. yodel grew per-distribution bounds. *(This bullet said `+Inf` until commit 7 — see below.)*
- `lease_lost` has a tile so the outcomes add up; cached months have a result so cache effectiveness is visible; archive errors are counted, since `ChessClient` maps only 404 to empty and throws otherwise — 429s and 5xx were invisible. Empty months no longer feed 0 into games-per-month. Motif labels lowercase with `Locale.ROOT`.

## Tests — where the panel was most useful

- **`TestOneD4Queries_GoldenInstrumentNames` pinned almost nothing.** Substring containment over a joined blob: a prefix rename of every instrument passed, stripping `service_name` from all seventeen selectors passed, deleting every timeseries passed. The first is the exact failure its own comment claimed to prevent. Rewritten per selector on the idiom this file already uses for portrait, name set closed both ways.
- **The disk test from commit 3 matched commented-out lines and prefix typos** — `root_path: /hostfs-typo` passed, which is the state its own error message describes. Anchored on active lines.
- Bucket placement was unpinned (everything in bucket 0 passed). Every metrics test indexed one month, so `monthCount` and `totalIndexed` were indistinguishable. `interrupted` and `lease_lost` had **no test at all** — the wedge alarm was the one outcome uncovered on both sides. The motif label was pinned with one detector registered, so a hardcoded `"check"` passed. "One pipeline per process" was true and untested.

## Corrections to my own claims

Commit 3's message said "three of the five scrapers"; there are six. This PR previously said every emit site was mutation-checked — **false**, four survived. Both corrected.

One mutation is deliberately left alive: removing `@Singleton` from the factory method does not change the observed instance under Micronaut's defaults, so it is equivalent, not a gap. The test pins the property, not the annotation.

---

# 5 — the cursor[bot] round

## The blocker

Commit 3's filesystem scraper used `match_type: glob`. **filterset accepts only `strict` and `regexp`** (`internal/filter/filterset/config.go`, v0.155.0) — an unrecognised type is an error out of `CreateFilterSet`, which fails the `hostmetrics` receiver, and with it the `otlp` pipeline it shares. Deploying that takes *every* metric down, not just the panels it was meant to fix.

The same block was wrong a second way: the excludes were written `/hostfs/var/lib/docker/*`, but `root_path` is applied when the scraper reads usage, not when it filters, so a `/hostfs`-prefixed pattern matches nothing.

Both fixed and both pinned — one test asserts the config only uses match types the collector accepts, another that mount-point patterns are unprefixed. cursor's stronger suggestion, a dry-run against the pinned image, I declined at the time and have since done — commit 8, which also confirms this blocker empirically.

## The rest

The host-wide `process` scraper is dropped. Nothing in prom_proxy reads it and it emits a series set per process on the machine.

Everything else in this commit is the same shape: a place where commits 1–4 made an argument in a comment and left no test behind.

- **`index_run_duration_micros` is labelled by outcome**, and both averages over it select `outcome="completed"`. An interrupted run sat at the `MAX_RUN` ceiling by definition, so pooling it in made the tile read worst exactly when it was being used to judge how bad things were — the same reason the motifs-per-game tile was left out.
- **`run_failure_rate` names `failed|interrupted`** rather than negating `completed`. `lease_lost` is the fourth outcome and it is ordinary, so the negation put a floor under the failure line; it also opted every future outcome in by default. Pinned both ways: no query in the entry may select outcomes by exclusion.
- **`RUN_DURATION_BOUNDS` is pinned.** Commit 4 added the bounds and asserted `distributionCount == 1`, which holds just as well when every run lands in the overflow bucket of a 10ms histogram — the exact failure commit 4 said it was fixing. Deleting the `defineDistribution` call passed every one_d4 test.
- An empty month is counted, not averaged into games-per-month — and `GAMES_PER_MONTH` declares count-shaped bounds of its own (commit 6), so it no longer borrows the HTTP latency buckets that happened to fit.
- **`CustomMetrics` copies bounds at every crossing**, including the snapshot boundary — the one that could hand a caller the shared `BUCKET_BOUNDS` static. The `Distribution` constructor's clone is *gone*: with the snapshot copying, no test can tell its absence from its presence, and an unpinnable guard is one that decays.
- `OtlpHttpMetricsExporter.close()` flushes once however many owners call it.
- The filter carries no `@PreDestroy` **and is not `AutoCloseable`** (Micronaut closes those with no annotation at all). Commit 4 fixed the double close and explained it in a comment; a test now enforces it.
- The `.table-wrap--wide` rules are read off the stylesheet and asserted. The class-name check alone passed against a deleted rule, which is the whole bug. Needs `@types/node` for `node:fs`; npm also resynced two stale version ranges in the lockfile's mirror of `package.json`.

---

# 6 — the +Inf claim was wrong, and it was hiding something

Three comments in this PR said an all-overflow histogram reads `+Inf` at p95. **It does not.** When the rank falls in the `+Inf` bucket, `histogram_quantile` returns the upper bound of the highest *finite* bucket, so the tile answers a flat `10000` and goes on answering it. That is worse than the claim it replaces: the broken histogram reads like a fast service, not like a broken histogram. Corrected in `IndexWorker`, `CustomMetrics` and `IndexWorkerTest`.

Chasing it surfaced that **the HTTP family has the same defect the run-duration histogram had**, and had it before this PR. `HttpServerMetrics.BUCKET_BOUNDS` is the OTel SDK default set — shaped for *milliseconds* — and this emitter records *microseconds*. Top bound 10ms. `server_pal` (Rust) and `futility/otel` (C++) inherit the same defaults from their SDKs and also record microseconds; neither configures a view. So **every service's p95 tile is capped at 10ms**, and anything touching Postgres is past that routinely. The `avg_duration_us` tile beside it is correct, since `_sum`/`_count` are unaffected — the disagreement between the two is how someone would eventually notice.

**Filed as #1286, not fixed here.** The three emitters have to move together or Java services stop being comparable to the rest on a query they share, and a bucket-layout change is not retroactive — both belong in their own diff. What is fixed here is the trap: a test pins Java's bounds equal to the OTel defaults and fails with the reason if they move alone (mutation-checked by shifting one bound). And `defineDistribution`'s javadoc no longer claims the default "suits anything request-shaped" — it does not suit that either.

---

# 7 — the dry-run I said I would not do

cursor's blocker came with a suggestion: *"a test that rejects `match_type: glob` and `/hostfs/`-prefixed mount points, **or better, a collector config dry-run against the pinned image**."* I did the first and declined the second, on the grounds that a docker-gated test skips silently wherever docker is absent — the failure mode this repo already has with the Postgres suites. That objection turns out to be answerable rather than disqualifying.

`scripts/validate-otel-config` starts the pinned collector against the real config and asserts it stays up. **Not `otelcol validate`** — that unmarshals the config and calls `Validate()` on it, which is weaker than building components, and the filterset error comes out of `CreateFilterSet` at scraper construction. The config has to actually start.

**It is self-validating**, which is what answers the silent-skip objection. Every run does the collector twice: once on the committed config expecting it to stay up, once on a copy with `match_type: glob` spliced back in expecting it to die. A docker that is missing, stubbed, or otherwise not discriminating fails one of the two halves, so the script cannot pass by doing nothing. Locally it was checked against three stub dockers — one that behaves (passes), one that always succeeds (fails the negative control), one that always fails (fails the positive). A missing binary returns 127 through `timeout`, not the 124 that means healthy, so absence is loud rather than quiet.

**Its first real run, on `5bd7e68`, passed and proved itself:**

```
otel-collector config dry-run against otel/opentelemetry-collector-contrib:0.155.0
  ok    the committed config builds its pipelines and stays up
  ok    match_type: glob is rejected by the collector (down:1)
```

The healthy run held the full 15s window; the `glob` copy died in 0.27s with exit 1. So the negative control fired against the actual pinned image — which independently confirms the blocker cursor found, rather than taking my reading of the source for it.

The image tag is read out of `docker-compose.observability.yml` rather than hardcoded, so the dry-run cannot drift onto a collector nobody runs.

It gets **its own CI job**, not a step in `test-deploy` — that one runs inside the `ci-base` container and this needs the runner's own docker. Path-gated on the three files that can change the answer.

`TestCIStillRunsTheCollectorDryRun` pins the wiring, since a workflow that stopped invoking the script would leave every other test in that file green. Exact line match rather than `strings.Contains`: containment survives a typo'd path, which is precisely how the golden-instrument test in this same PR managed to pin nothing.

## Mutation checks

Both `IndexWorker` pins, all three `CustomMetrics` copy sites, the exporter's `compareAndSet`, an injected `@PreDestroy` on the filter, the outcome label on the duration histogram, the empty-month guard, both prom_proxy selector changes, three edits to `index.css`, the HTTP-bounds guard, and the CI-wiring pin three ways (step commented out, path typo'd, `+x` removed). **All killed.**

**Two survived on the way there.** The first snapshot-aliasing assertion raised the *top* bound, which cannot move an observation to a different bucket — it passed against the aliased version. And the constructor clone had no observable effect at all, which is why it was removed rather than pinned.

---

## Verification

- **62/62** across `one_d4`, `yodel`, `prom_proxy`, `http_client` and `deploy/consolidated` with `--nocache_test_results`, against a **real Postgres 16.13**; PG-gated suites confirmed non-skipping from the JUnit XML. The cursor round re-ran **61/61** over the four packages it touches; commit 7 re-ran **59/59** over `yodel` and `one_d4`; commit 8 re-ran **61/61** over `deploy/consolidated`, `yodel`, `one_d4` and `prom_proxy`. The last three ran **without** `PG_TEST_DB_URL`, so the Postgres-gated one_d4 suites skipped there. CI runs them against the `postgres:18` service.
- **106/106** on `1d4_web`, plus `tsc --noEmit`. `buildifier_test` green, and no `MODULE.bazel.lock` churn.
- The new `validate-otel-config` job is green on `5bd7e68`, with its negative control firing — see above.
- Three context-booting `one_d4` suites needed `yodel:micronaut` in their runtime deps — the integration the binary already carries.
- **The three-lens panel did not run against commits 7 or 8.** Commit 7 is three comment corrections and one guard test; commit 8 is a CI script plus its wiring pin. Both are mutation-checked; neither got a panel, and I would rather say so than let the header imply otherwise.

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -28,6 +28,40 @@ jobs:
       - name: Test deploy.sh
         run: ./scripts/test-deploy
 
+  # deploy_config_test pins the text of the collector config, which can only
+  # catch mistakes someone already knew to look for: `match_type: glob` passed
+  # review and passed the line-presence test that existed at the time, and would
+  # have taken the whole metrics pipeline down on deploy. This starts the pinned
+  # collector against the real config instead.
+  #
+  # Its own job, not a step in test-deploy: that one runs inside the ci-base
+  # container and this needs the runner's own docker. Path-gated because it
+  # pulls the collector image, and nothing outside these three files can change
+  # the answer.
+  validate-otel-config:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Check for collector config changes
+        id: filter
+        run: |
+          CHANGED=$(git diff --name-only origin/main...HEAD -- \
+            deploy/consolidated/o11y/otel-collector.yml \
+            deploy/consolidated/docker-compose.observability.yml \
+            scripts/validate-otel-config)
+          if [ -n "$CHANGED" ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Dry-run the collector config
+        if: steps.filter.outputs.changed == 'true'
+        run: ./scripts/validate-otel-config
+
   # deploy.sh runs from a developer machine, and macOS ships bash 3.2 as
   # /bin/bash — no associative arrays or case-conversion expansions. The Linux
   # job above runs bash 5 and cannot reproduce that, so run the same suite on a
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -53,3 +53,11 @@ config_setting(
         "@platforms//os:linux",
     ],
 )
+
+# Neither .github/ nor scripts/ has a BUILD file, so these belong to the root
+# package. deploy/consolidated's dry-run wiring test reads both, and a go_test
+# only sees what it declares — so the root package has to export them first.
+exports_files([
+    ".github/workflows/branch.yml",
+    "scripts/validate-otel-config",
+])
--- a/deploy/consolidated/BUILD.bazel
+++ b/deploy/consolidated/BUILD.bazel
@@ -10,5 +10,7 @@ go_test(
     data = [
         "Caddyfile",
         "compose.yaml",
+        "docker-compose.observability.yml",
+        "o11y/otel-collector.yml",
     ],
 )
--- a/deploy/consolidated/BUILD.bazel
+++ b/deploy/consolidated/BUILD.bazel
@@ -12,5 +12,9 @@ go_test(
         "compose.yaml",
         "docker-compose.observability.yml",
         "o11y/otel-collector.yml",
+        # Read by TestCIStillRunsTheCollectorDryRun, which checks that CI has not
+        # quietly stopped invoking the collector dry-run.
+        "//:.github/workflows/branch.yml",
+        "//:scripts/validate-otel-config",
     ],
 )
--- a/deploy/consolidated/deploy_config_test.go
+++ b/deploy/consolidated/deploy_config_test.go
@@ -164,24 +164,61 @@ func readConfig(t *testing.T, name string) string {
 const hostfsMountPath = "/hostfs"
 
 func TestHostMetricsReadsTheHostAndNotTheCollectorContainer(t *testing.T) {
-	receiver := readConfig(t, "o11y/otel-collector.yml")
-	compose := readConfig(t, "docker-compose.observability.yml")
+	receiver := activeLines(t, "o11y/otel-collector.yml")
+	compose := activeLines(t, "docker-compose.observability.yml")
 
-	if !strings.Contains(receiver, "root_path: "+hostfsMountPath) {
-		t.Errorf("hostmetrics has no root_path: %s — every scraper measures the collector container,"+
-			" and the disk panels silently report nothing", hostfsMountPath)
+	// Exact lines, not substrings. `strings.Contains` would accept a commented-out
+	// directive and — worse — `root_path: /hostfs-typo`, which is the very state the
+	// error message below describes: a root_path naming a path the container has no
+	// mount for. Both were true of the first version of this test.
+	if !hasLine(receiver, "root_path: "+hostfsMountPath) {
+		t.Errorf("hostmetrics has no active `root_path: %s` — every scraper measures the collector"+
+			" container, and the disk panels silently report nothing", hostfsMountPath)
 	}
-	if !strings.Contains(compose, "- /:"+hostfsMountPath+":ro") {
-		t.Errorf("otelcol does not bind the host root at %s — root_path points at a path that does not"+
-			" exist in the container", hostfsMountPath)
+	if !hasLine(compose, "- /:"+hostfsMountPath+":ro") {
+		t.Errorf("otelcol does not bind the host root at %s — root_path points at a path that does"+
+			" not exist in the container", hostfsMountPath)
 	}
 
-	// Both scrapers are the ones that need the mount; the panels are named
-	// after them. Enabled-but-unmounted was the bug, so enabled is worth
-	// asserting alongside the mount rather than assumed.
+	// Enabled-but-unmounted was the bug, so the scrapers are asserted alongside the
+	// mount rather than assumed. These two are the ones whose panels were empty.
 	for _, scraper := range []string{"disk:", "filesystem:"} {
-		if !strings.Contains(receiver, scraper) {
+		if !hasLinePrefix(receiver, scraper) {
 			t.Errorf("hostmetrics scraper %q is not enabled; the Host page charts it", scraper)
 		}
 	}
+}
+
+// Config lines with comments and blanks removed, so an assertion cannot be
+// satisfied by a directive somebody commented out.
+func activeLines(t *testing.T, name string) []string {
+	t.Helper()
+	var out []string
+	for _, line := range strings.Split(readConfig(t, name), "\n") {
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+		out = append(out, trimmed)
+	}
+	return out
+}
+
+func hasLine(lines []string, want string) bool {
+	for _, line := range lines {
+		if line == want {
+			return true
+		}
+	}
+	return false
+}
+
+// For a mapping key whose value may be inline (`disk: {}`) or nested.
+func hasLinePrefix(lines []string, want string) bool {
+	for _, line := range lines {
+		if strings.HasPrefix(line, want) {
+			return true
+		}
+	}
+	return false
 }
--- a/deploy/consolidated/deploy_config_test.go
+++ b/deploy/consolidated/deploy_config_test.go
@@ -265,3 +265,49 @@ func TestMountPointExcludesAreWrittenFromTheHostsPerspective(t *testing.T) {
 		}
 	}
 }
+
+// The two tests above read the config as text, which bounds what they can find
+// to mistakes someone already thought of. `match_type: glob` was not one of
+// those: it passed review and passed the line-presence test of the day, and it
+// would have failed the hostmetrics receiver on deploy.
+//
+// scripts/validate-otel-config is the check that does not need to know the
+// mistake in advance — it starts the pinned collector against the real config.
+// But it only runs if CI invokes it, and a workflow that stopped invoking it
+// would leave every test in this file green. So the wiring is pinned too.
+//
+// This asserts the call exists, not that it passed; the job's own exit code is
+// the real signal. What it rules out is the silent version, where the dry-run
+// is dropped and nothing anywhere goes red.
+func TestCIStillRunsTheCollectorDryRun(t *testing.T) {
+	const script = "scripts/validate-otel-config"
+	workflow := readConfig(t, "../../.github/workflows/branch.yml")
+
+	// An exact line, not a substring. `strings.Contains` would be satisfied by a
+	// commented-out step, by a path that merely starts with this one, and by the
+	// script's name appearing in a comment — which is how the golden-instrument
+	// test in this same PR managed to pin nothing at all.
+	invocation := "run: ./" + script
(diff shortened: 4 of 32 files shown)


aaylward:
All of it, in `85a300b` and `afaa2cc`. The blocker was real and worse than an empty panel — thank you.

**Blocker.** Confirmed against v0.155.0 before changing anything: `validMatchTypes = []MatchType{Regexp, Strict}`. Now `regexp`, with the `/hostfs` prefix dropped from the mount points. Pinned two ways — `TestHostMetricsFilterConfigIsOneTheCollectorAccepts` closes the set of match types the config may use, and `TestMountPointExcludesAreWrittenFromTheHostsPerspective` rejects a `/hostfs/`-prefixed pattern. Both mutation-checked. I did not add a dry-run against the pinned image; the tests are static, so a match type this collector rejects for some *other* reason would still get through.

**`process` scraper.** Dropped.

**`RUN_DURATION` bounds.** You were right that deleting the `defineDistribution` line passed everything. It now asserts `boundsFor(RUN_DURATION)` equals the constant *and* that an ordinary run is not in the overflow bucket. Deleting the line kills it.

**Minors.** `run_failure_rate` is `outcome=~"failed|interrupted"`, and the test now rejects `outcome!=`/`outcome!~` anywhere in the entry — the negation also opted every future outcome in by default, which is the part that would have bitten later. Empty months pinned at `distributionCount(GAMES_PER_MONTH) == 0`. Stale javadoc pointer fixed.

**Double-close.** Fair — it was argument-not-test. Both halves are now tested, separately, because they are separate problems. `close()` flushes once however many owners call it (`aSecondCloseSendsNothingFurther`). And the filter is pinned to declare no `@PreDestroy` **and** not be `AutoCloseable` — Micronaut closes those with no annotation at all, so removing the annotation alone left the door open. Reflective, since the failure is an annotation that is absent. Mutation-checked by injecting a `@PreDestroy` method.

**Nits.** All four:
- Bounds are copied at the snapshot boundary now. The `Distribution` constructor's clone is *gone* — with the snapshot copying, no test could tell its absence from its presence, and that is a guard that decays. My first assertion for this was itself vacuous: it raised the *top* bound, which cannot move an observation to a different bucket, so it passed against the aliased version. Mutation-checking caught it, reading it did not.
- `GAMES_PER_MONTH` declares count-shaped bounds.
- `avg_run_seconds_1h` — fixed at the source rather than the query: the histogram is labelled by outcome, and both averages select `outcome="completed"`.
- CSS test now reads `index.css` and asserts the four rules the class hangs on. Deleting the `min-width` fails it. Still not a layout assertion — jsdom has none to give.

Two mutations survived on the way through, both mine, both fixed. Neither was visible by reading.

---

Issue #1287, opened 2026-08-02
Live metrics read against portrait: #1286 confirmed in production, Scene complexity measures renders, and counter tiles need a count/rate toggle


aaylward:
Read prom_proxy's live output for `portrait` against `api.muchq.com`, drove real traffic through `POST /portrait/v1/trace`, and read it again. Three findings, one of them evidence for an issue that until now rested only on reading the SDKs — plus one request (§4) that came out of reading the registry afterward.

## Method

Portrait had been idle for the full 7-day window — all nine series flat zero, 0 non-zero points out of 103 — so there was nothing to read until traffic existed. Two batches, 16 requests, all HTTP 200:

- **11 light requests** across three distinct scenes (3/5/1 spheres, 2/3/1 lights, 240x180–320x240), with deliberate repeats so 8 of them would hit the render cache.
- **5 heavy requests**, all distinct so none could hit cache: 10 spheres, 3 lights, 800x600, `reflective: 0.85`. Each took 0.8–1.1s wall clock.

Portrait exports on the futility/otel 10s default; Prometheus scrapes at 15s.

## 1 — #1286 is real in production, and the tile is insensitive to reality

`sum(http_server_requests_total)` and the duration tiles, before and after the heavy batch:

| | after light batch | after heavy batch |
|---|---|---|
| `avg_duration_microseconds` | 8,669.5 | **141,570.9** |
| `p95_duration_microseconds` | **10,000** | **10,000** |

Actual latency moved ~16x. The p95 tile did not move at all — it reports a flat `10000` in both states, held stable across four consecutive scrapes in each.

That is the top finite bound of the OTel SDK default bucket set, which is shaped for **milliseconds**, being read by `histogram_quantile` because every observation lands in `+Inf`. `avg` is trustworthy here because `_sum`/`_count` are unaffected by bucket layout — the disagreement between the two tiles is exactly the symptom #1286 predicted someone would eventually notice.

**The rigorous form of the argument**, since a heavy-tailed distribution *can* legitimately have a mean above its 95th percentile: 5 of the 16 requests in the window took ~0.8–1.1s wall clock. **31% of observations demonstrably exceeded 10ms**, so a correct p95 cannot be 10ms, whatever the shape of the tail.

[...]

Issue #1286, opened 2026-08-02
HTTP latency histograms record microseconds against millisecond-shaped buckets, so every p95 tile is capped at 10ms


aaylward:
Found while reviewing #1285, which fixed the same class of bug for one instrument (`index_run_duration_micros`). The HTTP family has it too, in all three languages, and it predates that PR.

## The mismatch

Every emitter records **microseconds** into a histogram whose bucket boundaries are the OTel SDK defaults:

```
0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000
```

Those defaults are unitless in the spec, but they are shaped for **milliseconds** — the top bound is meant to be 10 seconds. Read as microseconds, the top bound is **10 ms**.

| | site | how the bounds are set |
|---|---|---|
| Java | `HttpServerMetrics.BUCKET_BOUNDS` (`domains/platform/libs/yodel/.../HttpServerMetrics.java:22`) | hardcoded copy of the defaults |
| Rust | `server_pal/src/lib.rs:192`, `f64_histogram("http_server_request_duration_microseconds")` | no view — SDK default |
| C++ | `futility/otel/metrics.cc:62` `RecordLatency`, name suffixed `_microseconds` | no view — SDK default |

All three record µs. None configures a view. So the ceiling is uniform and uniformly wrong.

## What it does to the dashboard

`prom_proxy` charts p95 off those buckets for every service (`registry.go:198`, `registry.go:215`):

```promql
histogram_quantile(0.95, sum by (le) (rate(http_server_request_duration_microseconds_bucket{service_name="..."}[5m])))
```

Any request slower than 10 ms lands in the `+Inf` bucket. When the rank falls there, `histogram_quantile` **returns the upper bound of the highest finite bucket** — it does not return `+Inf`. So the p95 tile answers exactly `10000` and keeps answering it, no matter how slow the service gets.

That is the part worth fixing: the failure mode is not a visibly broken chart, it is a plausible one. A service sitting at a real p95 of 400 ms reports 10 ms, and the number is stable enough to look measured.

Anything doing a Postgres round trip is over 10 ms routinely, so this likely affects most services with a p95 tile today. `_sum` and `_count` are unaffected, so the `avg_duration_us` tiles beside it are correct — which makes the disagreement between the two tiles the most likely way someone would notice.

## Fix sketch

[...]

Issue #1276, opened 2026-07-31
golf_hub: TwoInstancesShareOneGame hangs ~60s waiting on cross-instance NOTIFY (2nd sighting, postgres idle throughout)


aaylward:
The 60s stall that `pg_hub_e2e_test`'s `size = "small"` ceiling was kept tight to catch has recurred. This is sighting #2; the first was on #1241.

That BUILD comment made the right call, and it worked:

```
# Small on purpose, even after the one-off 60s stall observed on
# #1241 (never reproduced across 40 local runs on postgres 16 and
# 18; the whole suite runs in ~2s): the tight ceiling is what turns
# a recurrence into a loud red instead of a quietly slow pass.
```

Seen on PR #1275's `build-and-test`, [run 30660499189](https://github.com/muchq/MoonBase/actions/runs/30660499189/job/91255353335). Nothing in that PR touches golf_hub or pg — it edits `.bazelrc`, which invalidates the graph and makes `diff-build` run the full 136-test suite. That is probably the main reason this is only the second sighting rather than the twentieth (see "exposure" below).

## It is one specific test, not general slowness

Five tests pass in ~276ms, then the sixth hangs for the full 59s:

```
[       OK ] PgGolfHubFixture.LiveGameSurvivesARestart (63 ms)
[       OK ] PgGolfHubFixture.StatsSurviveARestart (61 ms)
[       OK ] PgGolfHubFixture.PendingGameLifecycleWritesThrough (39 ms)
[       OK ] PgGolfHubFixture.CorruptRowsLoseTheGameNotTheLobby (61 ms)
[       OK ] PgGolfHubFixture.ConnectedFlagFollowsPresence (52 ms)
[ RUN      ] PgGolfHubFixture.TwoInstancesShareOneGame
I0000 ... hub_handler.cc:209] hub restored 0 rooms, 0 members, 0 games
I0000 ... hub_handler.cc:209] hub restored 0 rooms, 0 members, 0 games
-- Test timed out at 2026-07-31 20:38:16 UTC --
```

Both hub instances start fine. The hang is after that, in the part that depends on one instance seeing the other's write.

## Postgres was idle for the entire hang

This is the part that rules out the boring explanations. The service container's log covering the hang window (test ran 20:37:17 → 20:38:16):

```
20:35:49.130 [4611] FATAL: terminating connection due to administrator command   <- pg lib tests, before
20:39:19.934 [73]   LOG: checkpoint starting: time                               <- after
```

[...]

PR #1288, opened 2026-08-02
#1285 re-review follow-ups, then #1287: microsecond histogram buckets across all three rails, cache-hit scene complexity, and a count/rate toggle

aaylward:
Two commits. The first is the four small follow-ups this PR opened with; the second implements #1287 and is much the larger of the two.

> **Correction to `637ef31`'s commit message.** That message claims the Rust rail could not be compiled in the authoring sandbox. **That is wrong** — `//domains/platform/libs/server_pal:server_pal` builds, and `posterize`'s `rust_test`, which links it, passes. The mistake: an early bazel failure was actually disk exhaustion and got misattributed to `crate_universe` not fetching through the proxy, and a later `cargo check` failure on rustc 1.94.1 vs the workspace's required 1.97.1 is a *host cargo* limitation that has nothing to do with the build, since bazel uses rules_rust's own pinned toolchain. The commit message is left as pushed rather than force-rewritten under a running CI job; this note is the correction. Everything below reflects what was actually verified.

---

# Commit 2 — #1287 (`637ef31`)

Three of that issue's four sections. §3 is a deploy roll, not code.

## §1 — the histogram buckets

This is #1286's fix, landing in all three rails at once because prom_proxy runs one `histogram_quantile` expression across Java, Rust and C++ services and it only compares like with like when the bucket layouts match.

Every emitter recorded **microseconds** into the OTel SDK's default bounds, which are shaped for **milliseconds**. The top finite bound therefore meant 10ms; anything slower landed in `+Inf`; and `histogram_quantile` answers the highest finite bound when the rank falls there. So p95 read a flat `10000` forever, and held steady while doing it — #1287 caught it in production on portrait at a real p95 near a second.

New layout is 100µs → 10s. Java sets its array directly. **C++ and Rust had no bounds mechanism at all** — neither SDK lets you pass boundaries to the histogram factory — so both now register views. Both match on the `_microseconds` suffix rather than the single HTTP instrument name: `RecordLatency` appends that suffix to whatever it is handed and is the only producer of it, so portrait's `trace_request_duration_microseconds` had exactly the same defect and the same 10ms cap. `RecordDistribution` keeps its bare name and is deliberately not caught.

**A consequence worth calling out.** yodel's `CustomMetrics` fell back to `HttpServerMetrics.BUCKET_BOUNDS` for undeclared distributions. Reshaping that array for microsecond latency would have quietly made the fallback *worse* than what it replaced — against bounds starting at 100, every distribution over small counts collapses into bucket 0. `CustomMetrics.DEFAULT_BOUNDS` now holds the old SDK defaults, and a test pins the two as **not** equal, since re-merging them reads like a tidy-up and would pass every other test in the file.

`//domains/platform/libs/otel_contract` is the drift guard #1286 asked for. It reads the declaration out of all three sources and asserts they match — and separately that the shared layout is microsecond-shaped rather than the SDK defaults, because equality alone is satisfied by all three reverting together.

## §2 — scene complexity measured renders, not requests

`tracer_service.cc` returned on a cache hit before reaching the `RecordDistribution` calls, so `avg_spheres_1h` was a mean over *renders* under a label that read like a mean over *requests* — drifting further from offered load as the hit rate climbed. At portrait's observed 50% hit rate the panel described half the traffic.

Both paths now record, labelled by `cache_hit`, matching what `RecordLatency` already did in the same function. The registry splits the tiles into `requested` and `rendered` so both readings exist and are named.

The test uses a workload whose two means actually differ — 8 spheres over 4 requests (2.0) against 6 over 2 renders (3.0). #1287 notes that its own first batch couldn't separate the readings, since 33/11 and 9/3 are both 3.0; a test built on a homogeneous workload would have the same blind spot.

## §4 — count/rate toggle on counter tiles

`?view=count|rate`. A descriptor now carries a bare selector and the proxy wraps it in `increase()` or `rate()`, so the two forms cannot drift onto different series. Four counters that spent two tiles each on one series collapse to one.

**Cumulative is not offered.** `sum(x)` resets when its process does, so it reads low after every deploy — which is §3 of that same issue, an image that hadn't rolled. The 21 tiles that read that way are now `increase()` over a window.

Three alarm counters — `runs_failed`, `runs_interrupted`, `runs_lease_lost` — plus chat `failures` count over **24h** instead of 5m. A tile whose purpose is to be non-zero after something went wrong must not disarm itself between the failure and someone looking at the dashboard.

`StandardMetrics` is not view-driven: its JSON keys are a fixed struct the UI reads by name, and it already carries both forms. `requests_total` did convert to a windowed count, keeping its name so the UI — which lives in another repo — keeps rendering it. A golden-string test is the only thing that now says which of the two that field holds.

## Verification

- **Six mutation checks, all killed**: `alarmWindow` collapsed onto the default; the count form switched to `rate`; `requests_total` reverted to cumulative; and one bound altered in *each* of the three rails, to prove the contract test genuinely reads all three rather than passing on a parse failure.
- The alarm-window test was **vacuous** in its first form — collapsing the constant left every assertion comparing against the collapsed value and still passing. Strengthened before the mutation would kill it.
- **`scripts/diff-build origin/main`: 44/44 pass**, covering prom_proxy, yodel, otel_contract, portrait, posterize, mithril, aura, futility/otel, one_d4 and mcpserver.
- **All three rails compile, Rust included.** `server_pal` builds to an rlib and `posterize:test_posterize` links and passes against it. See the correction at the top — an earlier draft of this description claimed otherwise.
- **No review panel ran against this commit.**

---

# Commit 1 — #1285 re-review follow-ups (`9128c9f`)

Four things, none of which blocked #1285 — three are cursor[bot]'s remaining non-blocking items from the re-review at `5bd7e68`, which arrived after that PR had merged; the fourth is a doc error found while looking at them.

**The dry-run could pass without validating anything.** `run_collector` wraps `docker run` in `timeout`, and `docker run` pulls on a cold cache. On a runner without the image, a pull slower than `WINDOW` gets killed at **124** — the exact code that here means *"the collector stayed up"*. The positive half then passes having never parsed the config; the mutant run that follows finds warm layers, starts, dies on `glob`, and the negative control passes too. Both halves green, nothing proved — the self-validating property failing in the one place it is load-bearing. The pull now happens once, before anything is timed, and a failed pull exits non-zero. *This is the one change here that could not be exercised locally: docker client, no daemon.*

**Mount-point excludes were unanchored.** filterset's `regexp` matcher is Go's `MatchString`, which looks for the pattern anywhere in the string, so `/dev/.*` also excluded `/mnt/dev/data` — a real volume silently dropped off the usage panel. Anchored with `^` and pinned by `TestMountPointExcludesAreAnchored`, which fatals if it finds no patterns at all so it cannot pass by matching nothing.

**`run_failure_rate` was pinned only against exclusion.** The check forbade `outcome!=` and `outcome!~`, leaving two routes back to a wrong failure line that use no negation: widening to `failed|interrupted|lease_lost`, or dropping the selector for a bare total. The set is now asserted positively.

**`RecordDistribution` was attributed to the wrong language.** `CustomMetrics`' javadoc had it as server_pal's; it is futility/otel's, the same C++ library as the counters named beside it. The sentence also claimed the Rust rail already had the equivalent, and it does not — which is why `mithril` and `posterize` are still empty registry entries.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,7 +25,7 @@ traps that only surface in CI), [`docs/BUILD_AND_IDE.md`](docs/BUILD_AND_IDE.md)
 ## Commands
 
 ```bash
-scripts/diff-build origin/main   # what CI's build-and-test runs
+scripts/diff-build origin/main   # what CI's build-and-test runs — COMMIT FIRST, see below
 scripts/format-all               # bazel + java + cc + scala formatters
 scripts/mutation-check -f FILE -t 'TEST CMD' 's/OLD/NEW/'
 bazel test //domains/<path>/...
@@ -36,6 +36,12 @@ Bazel formatting do.
 
 ## Things that bite
 
+- **`scripts/diff-build` destroys uncommitted work.** It runs
+  `git checkout <base> --force` and back to hash both revisions, and `--force`
+  discards every modification to a tracked file. Untracked files survive, which
+  makes the loss look partial and easy to misread. There is no prompt and no
+  stash. Commit before running it, or run `bazel test` on the affected targets
+  instead — that is the same build without the checkout.
 - **`BUILD.bazel` `srcs` are listed by name**, with two globbing exceptions
   (`yochat_lib`, `wordchains_ios`). Elsewhere a new file not listed doesn't
   compile under Bazel and its tests don't run — while `go test ./...` or
--- a/deploy/consolidated/deploy_config_test.go
+++ b/deploy/consolidated/deploy_config_test.go
@@ -257,12 +257,42 @@ func TestMountPointExcludesAreWrittenFromTheHostsPerspective(t *testing.T) {
 		case strings.HasPrefix(line, "exclude_fs_types:"), strings.HasPrefix(line, "network:"):
 			inExcludeBlock = false
 		case inExcludeBlock && strings.HasPrefix(line, "- "):
-			pattern := strings.TrimPrefix(line, "- ")
+			// The patterns are anchored (see the test below), so compare the path part.
+			pattern := strings.TrimPrefix(strings.TrimPrefix(line, "- "), "^")
 			if strings.HasPrefix(pattern, hostfsMountPath+"/") {
 				t.Errorf("mount-point exclude %q is prefixed with %s; filtering happens before"+
 					" root_path is applied, so this can never match", pattern, hostfsMountPath)
 			}
 		}
+	}
+}
+
+// filterset's regexp matcher is Go's MatchString, which looks for the pattern
+// anywhere in the string rather than at the front. These excludes are all meant
+// as path prefixes, and unanchored they overreach: `/dev/.*` drops `/mnt/dev/data`
+// too. The panel still renders, just missing a volume nobody thought to exclude —
+// the same shape of quiet wrongness as the /hostfs prefix above.
+func TestMountPointExcludesAreAnchored(t *testing.T) {
+	inExcludeBlock := false
+	seen := 0
+	for _, line := range activeLines(t, "o11y/otel-collector.yml") {
+		switch {
+		case strings.HasPrefix(line, "exclude_mount_points:"):
+			inExcludeBlock = true
+		case strings.HasPrefix(line, "exclude_fs_types:"), strings.HasPrefix(line, "network:"):
+			inExcludeBlock = false
+		case inExcludeBlock && strings.HasPrefix(line, "- "):
+			seen++
+			pattern := strings.TrimPrefix(line, "- ")
+			if !strings.HasPrefix(pattern, "^") {
+				t.Errorf("mount-point exclude %q is unanchored; filterset matches it anywhere in"+
+					" the path, so it also excludes mounts that merely contain it", pattern)
+			}
+		}
+	}
+	// Without this the test passes just as well against a deleted exclude block.
+	if seen == 0 {
+		t.Fatal("found no mount-point excludes to check; this test is not reading the config")
 	}
 }
 
--- a/deploy/consolidated/o11y/otel-collector.yml
+++ b/deploy/consolidated/o11y/otel-collector.yml
@@ -39,15 +39,21 @@ receivers:
         # Paths are the host's, unprefixed: root_path is applied when the scraper
         # reads usage, not when it filters, so a /hostfs-prefixed pattern matches
         # nothing.
+        #
+        # Anchored with ^. filterset's regexp matcher is Go's MatchString, which
+        # looks for the pattern anywhere in the string, so an unanchored /dev/.*
+        # also excludes /mnt/dev/data — a real volume, dropped off the usage
+        # panel because its path happens to contain one of these. These are meant
+        # as prefixes, so they say so.
         exclude_mount_points:
           mount_points:
-            - /var/lib/docker/.*
-            - /var/lib/kubelet/.*
-            - /snap/.*
-            - /run/.*
-            - /dev/.*
-            - /sys/.*
-            - /proc/.*
+            - ^/var/lib/docker/.*
+            - ^/var/lib/kubelet/.*
+            - ^/snap/.*
+            - ^/run/.*
+            - ^/dev/.*
+            - ^/sys/.*
+            - ^/proc/.*
           match_type: regexp
         exclude_fs_types:
           fs_types:
--- a/domains/graphics/apis/portrait/tracer_service.cc
+++ b/domains/graphics/apis/portrait/tracer_service.cc
@@ -1,9 +1,11 @@
 #include "domains/graphics/apis/portrait/tracer_service.h"
 
 #include <exception>
+#include <map>
 #include <memory>
 #include <new>
 #include <optional>
+#include <string>
 #include <vector>
 
 #include "absl/log/log.h"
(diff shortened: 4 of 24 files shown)


aaylward:
Thanks — finding 1 was right and was the one I'd have regretted. Fixed in `eeda52a`, along with 2 and 3. Answering 4 and 5b rather than changing them, with reasons.

## 1 — view wiring, now pinned on both rails

You were exactly right that the constant was pinned and the thing that *applies* it was not.

**C++ is fully closed.** `otel_provider_test.cc` drives a real `MeterProvider` through opentelemetry-cpp's in-memory exporter and reads `boundaries_` back off an exported histogram. Four mutations, all killed:

| mutation | killed by |
|---|---|
| predicate → `.*_nanoseconds` | the two positive tests |
| predicate → `.*` (every histogram) | `ADistributionKeepsTheDefaultLayout` |
| predicate → `http_server_.*_microseconds` | `PortraitsRenderTimerIsCoveredByTheSamePredicate` |
| `RegisterLatencyBucketView(*meter_provider);` deleted from the ctor | `TheConstructorRegistersTheView` |

That last row is the one your finding turned on, and it's why the file has two kinds of test. Four of them call the view function directly; a fifth constructs `OtelProvider` and reads the view off the provider it publishes globally, because a test that calls the function itself leaves the *call site* free to be deleted. There's also a control asserting that **without** the view the same instrument gets the SDK defaults — otherwise the positive assertions could be passing for some reason other than the view. `RegisterLatencyBucketView` moved out of the anonymous namespace to make this reachable.

**Rust is closed at the predicate only, and I want to be explicit that it's weaker.** `Instrument` has no public constructor, so the view closure can't be driven from a test — your suggestion was the right shape and it's what I did: `wants_latency_buckets` is extracted and tested, three mutations killed including `ends_with` relaxed to `contains`. But the `.with_view(latency_buckets)` call site on that rail stays unpinned. Deleting that line still leaves Rust green. Not equivalent to C++, and I'd rather say so than let the symmetry of "both rails now have tests" imply otherwise.

**A side effect worth flagging:** adding the `rust_test` target revealed that `lib.rs` already had a `#[cfg(test)]` module nothing had ever compiled — no test target existed for the crate, so its two rate-limiter tests had never run under Bazel. `cargo test` would have run them; CI doesn't use cargo. They run now, and that's what surfaced the missing `tower` dev-dependency in the graph.

## 2 — alarm window floor

Correct, `"6m"` passed. Now parses the constant and requires `>= 1h`; mutation-checked at `6m`, killed.

## 3 — JSON keys

Correct, and the same reasoning I'd already applied to `requests_total`'s value. Now asserted on the raw payload: `"view"` present, a counter tile carries `"toggleable": true`, and a fixed-form tile omits the key entirely rather than sending `false` — the negative half matters because `omitempty` is what tells the UI not to draw a toggle it can't honour.

## 4 — timeseries key renames

Checked rather than coordinated, and I think these are safe. Two comments already in this repo describe how the UI treats custom series, both pointing the same way: `registry.go` says a custom key colliding with `STANDARD_SERIES` "is classified as standard by the UI and silently never charts", and `service_handlers.go` says "the UI's generic Trends grid renders custom series in payload order with position-indexed colors". Both describe custom keys as handled generically by exclusion from a standard list, not looked up by name. So a rename changes a chart's label, not whether it renders.

I can't verify that directly — muchq.github.io isn't in this session's repo scope — so it rests on those two comments being accurate. Flagging it here so it's a decision rather than an assumption.

## 5b — `scene_*` unscoped by `service_name`

This one I looked into and **the suggested change would break the tiles**, which is worth recording.

`MetricsRecorder` takes `service_name` and uses it as the *meter* name (`metrics.cc:33`), not as an attribute — it never attaches it to a recorded point. The reason `cache_hits_total` carries the label is that `aura::Cache` adds it explicitly, from `metrics_->service_name()` (`cache.h:48`). `RecordDistribution` passes through only what the caller hands it, which for scene complexity is now just `cache_hit`.

So `scene_sphere_count` has no `service_name` label at all, and adding `service_name="portrait"` to those queries would select nothing and blank the panel. The asymmetry you noticed is real, but it's in the emitter, not the query — either the recorder should attach `service_name` to everything it records, or these series can't be scoped. That's a genuine follow-up and bigger than this PR; happy to file it.

## 5a, 5c, 5d — waived, briefly

- **Alarm tiles toggleable → `rate(...[24h])`.** Valid, low-utility, kept for uniformity. It isn't currently pinned as intentional; I left it that way rather than adding a test that asserts a shrug.
- **`docker pull` line order not text-pinned.** Fair. From commit 1 and I didn't reopen it.
- **Batching three things into one PR.** Also fair, and the working agreement does prefer one item per PR. #1287 was asked for on this branch explicitly; the `diff-build` warning rode along because I hit that footgun while doing the work. I'd split it differently given the choice again.

20/20 green across prom_proxy, futility/otel, server_pal, otel_contract, yodel, portrait and aura. No panel ran against `eeda52a` either.

---

aaylward:
`build-and-test` went red on `eeda52a` — **#1276, sighting #3**, not this diff. Re-triggered with an empty commit (`33dd6c8`); the rerun-failed-jobs API isn't available to this token. Full write-up on the issue: #1276 (comment).

Short version: `PgGolfHubFixture.TwoInstancesShareOneGame` timed out at its small ceiling, the one test in that fixture that waits on a cross-instance NOTIFY, with both hubs already logged up. Identical signature to the two prior sightings.

Why a metrics PR ran the golf_hub pg suites at all: `pg_hub_e2e_test` deps `//domains/platform/libs/futility/otel:metrics`, and this branch edits that package's `BUILD.bazel` (`exports_files` + the new `otel_provider_test` target). That re-hashes the package and fans out to every dependent of the shared metrics rail. `:metrics` itself has no source or dep change, and the otel edit is additive and confined to `otel_provider.{h,cc}` — the blast radius is package-level hashing, not the code.

That's the same shape as the other two sightings, both also on PRs that invalidated a wide slice of the graph and neither touching golf_hub.

Not fixing it here — the deadline-and-catch-up work #1276 asks for is a `golf_hub`/`pg:listener` change and doesn't belong in this PR.

---

PR #1289, opened 2026-08-02
Fix lost LISTEN/NOTIFY wakeup that hung TwoInstancesShareOneGame (#1276)

aaylward:
## Summary

Closes the intermittent `TwoInstancesShareOneGame` hang (#1276 / #1241): CI shows both hubs restore, then Bazel’s 60s `size = "small"` kill, with postgres idle. Do **not** raise the timeout.

Two cooperating failure modes share that signature:

1. **Lost wakeup in `pg::Listener`** — `PQexec(LISTEN/UNLISTEN)` can absorb a NOTIFY into libpq’s queue with no `POLLIN` left for `poll`.
2. **TearDown hang in `SharedViewOwner::End`** — unread wake projections leave a harvested async send waiter holding the pin while receive-first close ordering runs `~AsyncEventStream` → `End()` → `Close()` on the same stack. **This is a smithy-cpp defect** (fire send before receive in both transports); `QuiesceCrossTable` is the fixture workaround so the helper is not mystery scaffolding. Production trigger: peer close while a registry fan-out write is outstanding (golf_hub uses Beast + `async_delivery`).

## Fix

1. Drain `PQnotifies` after every `SyncChannels` pass.
2. Nonblocking wake pipe (`O_NONBLOCK` + `EAGAIN`-aware drain) so a full 64-byte read cannot hang the poll thread.
3. Room catch-up on `OnChannelActive` via shared `WakeChannel` / `CatchUpRoom`: Flush+Load **off `mu_`**, re-project **only when rows moved** (notify still always re-projects). Avoids reconnect/boot queue floods.
4. Named await deadlines; quiesce armed **before** `SeatedCrossTable`, drains seats on setup failure, covers chat cross-instance tests; pinned by `QuiesceDrainsUnreadWakeFrames`.
5. Hang-safe absorb regression test (release-on-unwind, bounded `wait_for`, no sleep; waits for `other` active to pin the LISTEN pass).
6. `Listener` requires `on_active` (pass `nullptr` explicitly for notify-only).
7. Keep `size = "small"`; BUILD comment states both hang shapes.

## Review follow-ups (`6b1371a`)

Addresses the panel on `6005605` / `deda1f5`: room catch-up storm, quiesce coverage gap, hang-prone absorb test, wake-pipe block, BUILD over-claim, unheld-room active guard test, merge of notify/active dispatch.

## Tests

| Check | Result |
|---|---|
| `listener_test` (incl. absorb + reconnect) | PASSED |
| `hub_store_race_test` (active refresh + unheld guard) | PASSED |
| `pg_hub_e2e_test` full | PASSED (~2s) |
| TwoInstances + QuiesceDrains + RemoteFinish ×20 | all pass, ~1.6s |

Sanitizers still exclude `golf_hub` / `pg` packages per `branch.yml`.

<div><a href="https://cursor.com/agents/bc-019fc3bc-defe-7b9b-8734-c3180586b027?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019fc3bc-defe-7b9b-8734-c3180586b027&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/domains/games/apis/golf_hub/BUILD.bazel
+++ b/domains/games/apis/golf_hub/BUILD.bazel
@@ -317,10 +317,13 @@ cc_library(
 # skips otherwise.
 cc_test(
     name = "pg_hub_e2e_test",
-    # Small on purpose: the 60s stall on TwoInstancesShareOneGame
-    # (#1241, #1276) is a lost wakeup, not marginal slowness — normal
-    # runtime is ~2s. The tight ceiling turns a recurrence into a loud
-    # red instead of a quietly slow pass; do not raise it.
+    # Small on purpose (#1241, #1276): normal runtime is ~2s; the tight
+    # ceiling turns a recurrence into a loud red. Two independent hang
+    # shapes share the signature — a lost LISTEN/NOTIFY wakeup (fixed in
+    # pg::Listener) and a TearDown End() deadlock when unread wake
+    # frames pin the async delivery chain (worked around here by
+    # QuiesceCrossTable; root cause is send-before-receive ordering in
+    # smithy-cpp). Do not raise the ceiling.
     size = "small",
     srcs = [
         "pg_hub_e2e_test.cc",
--- a/domains/games/apis/golf_hub/BUILD.bazel
+++ b/domains/games/apis/golf_hub/BUILD.bazel
@@ -317,10 +317,10 @@ cc_library(
 # skips otherwise.
 cc_test(
     name = "pg_hub_e2e_test",
-    # Small on purpose, even after the one-off 60s stall observed on
-    # #1241 (never reproduced across 40 local runs on postgres 16 and
-    # 18; the whole suite runs in ~2s): the tight ceiling is what turns
-    # a recurrence into a loud red instead of a quietly slow pass.
+    # Small on purpose: the 60s stall on TwoInstancesShareOneGame
+    # (#1241, #1276) is a lost wakeup, not marginal slowness — normal
+    # runtime is ~2s. The tight ceiling turns a recurrence into a loud
+    # red instead of a quietly slow pass; do not raise it.
     size = "small",
     srcs = [
         "pg_hub_e2e_test.cc",
--- a/domains/games/apis/golf_hub/hub_handler.cc
+++ b/domains/games/apis/golf_hub/hub_handler.cc
@@ -305,47 +305,55 @@ void HubHandler::AttachListener(pg::Listener* listener) {
 }
 
 void HubHandler::OnNotify(const std::string& channel, const std::string& payload) {
+  WakeChannel(channel, payload, /*from_active=*/false);
+}
+
+void HubHandler::OnChannelActive(const std::string& channel) {
+  WakeChannel(channel, /*payload=*/"", /*from_active=*/true);
+}
+
+void HubHandler::WakeChannel(const std::string& channel, const std::string& payload,
+                             bool from_active) {
   constexpr std::string_view kChatPrefix = "chat_";
   if (absl::StartsWith(channel, kChatPrefix)) {
     // Own wakes pump too, deliberately: the cursor makes it a no-op in
     // the common case, and it is what recovers an append whose
-    // connection died between COMMIT and the reply.
+    // connection died between COMMIT and the reply. Active has no
+    // payload, so the own-instance filter never applies here.
     PumpChat(std::string(channel.substr(kChatPrefix.size())));
     return;
   }
-  if (payload == instance_id_) return;  // our own commit; locals already heard it
+  // Notify echoes of our own commit are skipped; locals already heard
+  // them. Active signals carry no instance id — always a catch-up.
+  if (!from_active && payload == instance_id_) return;
   constexpr std::string_view kPrefix = "room_";
   if (!absl::StartsWith(channel, kPrefix)) return;
-  const std::string room_id = channel.substr(kPrefix.size());
-  Outbox outbox;
+  // Notify always re-projects (wake contract). Active only projects when
+  // rows moved — reconnect re-LISTENs every room and must not flood.
+  CatchUpRoom(std::string(channel.substr(kPrefix.size())), /*project_always=*/!from_active);
+}
+
+void HubHandler::CatchUpRoom(const std::string& room_id, bool project_always) {
   {
     const std::lock_guard<std::mutex> lock(mu_);
     // Only rooms we hold: a stale wake for a dropped room must not
     // resurrect it (join is the one path that materializes rooms).
     if (!rooms_.contains(room_id)) return;
-    RefreshRoomLocked(room_id, outbox);
   }
-  Deliver(outbox);
-}
-
-void HubHandler::OnChannelActive(const std::string& channel) {
-  // Rows committed while we were not subscribed queued no notification.
-  // Chat heals from its cursor; rooms re-read and re-project like a wake
-  // (#1276) — the notify-only path is not enough around (re)LISTEN.
-  constexpr std::string_view kChatPrefix = "chat_";
-  if (absl::StartsWith(channel, kChatPrefix)) {
-    PumpChat(std::string(channel.substr(kChatPrefix.size())));
+  // Off mu_: reconnect fires this once per room on the poll thread; holding
+  // the lock across Flush+LoadRoom would stall every move/chat/join for
+  // the whole burst. Same shape as PumpChat.
+  store_->Flush();
+  auto rows = store_->LoadRoom(room_id);
+  if (!rows.ok()) {
+    LOG(WARNING) << "room " << room_id << " catch-up failed: " << rows.status();
     return;
   }
-  constexpr std::string_view kPrefix = "room_";
-  if (!absl::StartsWith(channel, kPrefix)) return;
-  const std::string room_id = channel.substr(kPrefix.size());
   Outbox outbox;
   {
     const std::lock_guard<std::mutex> lock(mu_);
-    // Same membership guard as OnNotify: never materialize from a wake.
     if (!rooms_.contains(room_id)) return;
-    RefreshRoomLocked(room_id, outbox);
+    ReconcileRoomLocked(room_id, *rows, outbox, project_always);
   }
   Deliver(outbox);
 }
@@ -353,7 +361,8 @@ void HubHandler::OnChannelActive(const std::string& channel) {
 void HubHandler::RefreshRoomLocked(const std::string& room_id, Outbox& outbox) {
   // The flush means this read can never be older than our own truth;
   // holding mu_ across it means nothing local moves in between. The
-  // writer thread needs no lock we hold, so it drains freely.
+  // writer thread needs no lock we hold, so it drains freely. Join uses
+  // this path to materialize; notify/active use CatchUpRoom instead.
   store_->Flush();
   auto rows = store_->LoadRoom(room_id);
   if (!rows.ok()) {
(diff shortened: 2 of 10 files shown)


aaylward:
## Review panel — four lenses (correctness/control flow; libpq/locks/resource safety; tests/docs/CI; altitude)

Panel ran read-only against `6005605`. Every finding below I re-verified myself before reporting it; the ones I could not settle by reading are called out as such rather than asserted.

Two agents concluded the PR's **root cause #2 does not hold** — `SharedViewOwner::End()` closes the socket before waiting, and `Close()` fires parked sends, releasing the pin. **They were wrong, and the reason matters**, so that goes first.

---

## 1. `QuiesceCrossTable` is masking a production-reachable deadlock in smithy-cpp

Root cause #2 is real, the mechanism is more specific than the commit message says, and it is **not test-only**. Verified in the pinned tree (`v0.1.0`, and identical on smithy-cpp `main` @ `064b427`):

- `TakeAsyncWaitersLocked` (`beast_transport.cc:619`) `std::exchange`es **both** the parked receive and the parked send to null in one locked step.
- `CompleteAsyncWaiters` (`:625-639`) then fires **receive first, send second**. `PairEnd::Close` (`websocket_pair.cc:119-124`) has the same order.
- `InvokeCompletion` (`:83`) is a bare synchronous call, not a post — so the receive completion resumes the coroutine *on that stack*.
- The coroutine ends → `~AsyncEventStream` → `End()` → `socket_->Close()` finds `pending_send_` **already null**, because it was harvested a moment ago → `idle.wait(active == 0)`.
- The one outstanding pin belongs to the send callback in `waiters.send`, one frame below, which can only fire after the receive call returns. It never returns.

`SendAsync` releases its pin *inside* the send callback (`event_stream.h:183`), so "a send waiter is outstanding" and "a pin is held" are the same condition — the wait is unsatisfiable by construction on this path.

The refutation both agents relied on is correct **only for sends still parked in the socket**. Once harvested, `Close()` has nothing to fire, and `OnWrite` can't fire it either (`:683`, by design — `:842` says the terminal transition takes and fires the waiters exactly once).

Production trigger: a peer close frame or TCP reset arriving while a registry fan-out write is outstanding to that client — closing the tab mid-game. golf_hub runs exactly that configuration (`golf_hub_main.cc:207,214` — Beast, `async_delivery = true`). Each occurrence wedges one io thread.

**This is not #1289's bug to fix** — it is two lines in smithy-cpp (fire send before receive in both transports). But it changes what this PR *is*: the quiesce helper is a workaround for a live service defect, and nothing in the PR, the BUILD comment, or #1276 currently says so. Suggest landing it as-is with that stated plainly, so the helper isn't deleted later as mystery scaffolding. A smithy-cpp issue is written up but not yet filed — the App lacks issue-write on that repo.

**Caveat I want to be honest about:** this chain is derived by reading. I have no stack from a wedged process. The PR says gdb was used on hung runs — if that stack shows `SharedViewOwner::End` under `CompleteAsyncWaiters`, it confirms this outright and tells us whether the quiesce can eventually be deleted.

---

## 2. High — the room catch-up is a reconnect/boot storm in production

`hub_handler.cc:340-350` with `listener.cc:111-114`. Found independently by two lenses.

On reconnect `active_` is cleared, so `SyncChannels` fires `on_active_` for **every** re-LISTENed channel. Each `room_*` signal now runs `RefreshRoomLocked` — `store_->Flush()` (an unbounded `cv_` wait) plus `LoadRoom`, both DB round trips — **under the global `mu_`**, serialized on the single poll thread, then re-projects a `roomState` per member plus a `gameState` per game seat.

An instance holding 300 rooms that loses its listener connection does 300 serialized flush+read round trips holding `mu_`, blocking every move, chat and join for the duration, then floods per-session queues bounded at 64 whose default overflow policy is close-the-session. Clients that were slightly behind get disconnected *by the recovery path*. It fires precisely when the database is least healthy, and a flapping connection repeats it. The same burst runs at boot: `RestoreFromStore` reads the snapshot, then `AttachListener` immediately provokes a full per-room re-read.

Before this change, reconnect catch-up was chat-only, and `PumpChat` is a cursor no-op in the common case. Worth a dirty-room filter, or coalescing the active-signal refresh, or at minimum doing the loads off `mu_` the way `PumpChat` already does.

---

## 3. High — a pre-existing lost-wakeup that survives this PR

`listener.cc:157-161`. `pipe(fds)` at `:43` sets no `O_NONBLOCK`, and the drain is:

```cpp
char drain[64];
while (read(wake_read_, drain, sizeof(drain)) == sizeof(drain)) {
}
```

If the queued byte count is an exact multiple of 64, the last read returns 64, the loop iterates, and the next read **blocks forever** on an empty blocking pipe. The poll thread never returns to `poll()`, never calls `PQconsumeInput`, never drains notifies — a lost wakeup with postgres idle, which is the #1276 signature exactly. Reachable at boot, where `AttachListener` bursts two bytes per room while the poll thread is inside `PQconnectdb`.

Not what hit CI (these tests hold ~1 room), and not this PR's regression. Flagging it because **#1276 shouldn't be closed as fully explained while this stands** — it's the same symptom class from a different cause. Fix is `pipe2(fds, O_NONBLOCK)` or an `EAGAIN`-aware loop.

---

## 4. Medium — the new regression test hangs its own target instead of failing

`listener_test.cc:144`, `:157`, `:165`. All three lenses found this.

The callback parks the poll thread on `cv.wait(lock, [&]{ return release; })` with no deadline, and `release` is only set at `:169`. Two `ASSERT`s sit between the park and the release — notably `ASSERT_TRUE(client.Exec("SELECT pg_notify('ch','victim')").ok())`. If either fires, the test returns with `release == false`, `~Listener` joins a thread that will never leave the callback (`stopping_` is only read at the top of `Loop`), and the `small` target dies at 60s with no failure text.

A PR whose purpose is converting silent hangs into named failures has introduced a new silent hang on its own failure path. A `wait_for` with a generous bound plus a release-on-timeout fixes it.

---

## 5. Medium — the quiesce guard doesn't cover the path that actually fails

`pg_hub_e2e_test.cc:695` and `:770`:

```cpp
auto table = SeatedCrossTable(*remote);
ASSERT_TRUE(table.has_value());
QuiesceOnScopeExit quiesce{this, remote.get(), &*table};
```

`SeatedCrossTable` is where all six new deadlines live. A lost wake is now exactly what makes it return `nullopt` — so on the failure this PR exists to diagnose, `QuiesceOnScopeExit` is never constructed and teardown runs un-quiesced. The 5s named `ADD_FAILURE` goes into gtest's buffered stdout, teardown parks, Bazel SIGKILLs at 60s, and the buffered diagnosis dies with the process. Back to `[ RUN ]` and silence.

Note the guard takes `&*table`, so this isn't just moving the line up — it needs to hold an `optional<CrossTable>*`, or `SeatedCrossTable` needs to drain on its own failure exits. Cleaner still: give `CrossTable` a quiesce member declared **last**, so reverse-order destruction runs it before the seats and no test can forget it.

Same gap, smaller exposure: `ChatCrossesInstancesBothWays:832` and `ChatCommittedDuringListenerOutageArrivesAfterReconnect:895` build cross-instance seats and leave frames unread with no guard at all.

---

## 6. Medium — coverage gaps

- **`QuiesceCrossTable` is unpinned.** Delete both `QuiesceOnScopeExit` lines and the suite still passes. One of the two claimed root causes has no test at any level, and `×40 green` cannot distinguish "quiesce fixed teardown" from "the drain fix alone sufficed and quiesce is inert."
- **The membership guard's negative case is unpinned.** `hub_handler.cc:347` carries a comment stating a contract; the nearest test (`hub_chat_race_test.WakesForUnheldRoomsAreIgnored`) only fires *chat* channels, never `OnChannelActive(RoomChannel(...))` for an unheld room. Remove the guard and nothing fails.
- **`DrainsNotifyBufferedByListenExec` isn't deterministic.** A 50ms `sleep_for` is the only thing making the victim arrive before `SyncChannels`' `PQexec`. If it slips on a loaded runner (`branch.yml:180` caps at 4 vCPU while the whole C++ tree builds), the fixed code delivers via the ordinary POLLIN path and the assertions hold identically. It doesn't flake red — it flakes **green**, reporting coverage of a path that didn't run. Nothing distinguishes the two delivery paths.

---

## 7. Medium — the BUILD comment now over-claims

`BUILD.bazel:320-323` asserts the stall "is a lost wakeup, not marginal slowness." But `pg_hub_e2e_test.cc:256-258`, added in the same commit, documents a second independent cause that is not a lost wakeup, is not fixed in production code, and is worked around only in the fixture. A reader who trusts the BUILD comment concludes the stall is fully explained and fully closed. Dropping the "never reproduced across 40 local runs" line was right; the replacement goes too far.

---

## What checks out

The drain fix itself is correct and is the strongest part of this change:

- **Placement is sound.** `PQnotifies` parses already-buffered bytes without touching the socket, so the post-`SyncChannels` drain recovers exactly what `PQexec` absorbed, and anything still in the kernel buffer keeps POLLIN asserted and heals next pass — including a partial message straddling a recv boundary. No `PQconsumeInput` needed at that call site.
- **The pre-fix bug was total, not merely slow** — `PQnotifies` was reachable only from the POLLIN branch, and the 1000ms timeout loops back through a `SyncChannels` that issues no `PQexec` when nothing changed, so an absorbed notify was lost forever.
- `PQfreemem` pairs on every path; no leak.
- **No concurrent `PGconn` use.** The `PGconn` is poll-thread-only; `Listen`/`Unlisten` touch only `wanted_` under `mu_`.
- **The `OnChannelActive → RefreshRoomLocked` deadlock concern really is refuted** — independently re-derived. `SyncChannels` releases `Listener::mu_` before the `on_active_` loop, so the order is always `HubHandler::mu_ → Listener::mu_`, never the reverse. Your earlier panel got this right.
- **Catch-up can't double-apply or resurrect**: the `contains` guard blocks materialization, `if (over) continue` blocks a second ceremony, and `ListenRoomLocked` iterates a copy so re-entrant `Listen` can't invalidate the loop.
- Both new tests genuinely fail against `4eb1495`. BUILD wiring is complete, both pg gates have the right `env_inherit`, `size = "small"` is retained, and the deadlines compose under it (5s budget threaded as *remaining* time, so the total stays bounded).

**One CI note:** no sanitizer covers any file in this PR — `branch.yml:250-255` excludes both `//domains/games/apis/golf_hub/...` and `//domains/platform/libs/pg/...`. For a concurrency change that is worth stating in the PR body rather than leaving implicit.

---

## Smaller design points

- **`OnNotify` and `OnChannelActive` are now two copies of one contract** differing by one line (`hub_handler.cc:307-351`). A private `Wake(channel)` holding the chat/room dispatch collapses both, ≈−17 lines, and stops the next channel prefix being added to one path and forgotten in the other. The self-skip has to stay chat-exempt, so the merged guard needs writing as `if (payload == instance_id_ && !StartsWith(channel, kChatPrefix)) return;`.
- **`ActiveCallback on_active = nullptr` (`listener.h:37`) makes the lossy wiring the default** in a shared library — a caller passing only `on_notify` silently gets the bug this PR fixes, and four of the six `pg::Listener` constructions in the repo omit it. Dropping the default argument forces a visible choice.
- `AwaitRoomState`/`AwaitGameView` are now near-identical 22-line functions with duplicated failure strings; on timeout the caller also gets `ReceiveCase`'s uninformative "TimeoutError" printed *before* the new diagnostic.

---

## Needs a stack, not a reader

Whether the deadlock in §1 is the hang actually observed in CI. The mechanism is real either way, but only a stack from a wedged run decides whether `QuiesceCrossTable` can be deleted once smithy-cpp is fixed, or whether it's covering something else as well.

---

PR #1292, opened 2026-08-02
prom_proxy: count/rate as separate always-present series across the Serving and Trends charts

aaylward:
## Summary

- The dashboard's count/rate toggle (#1287) only ever affected custom scalar tiles — the Serving section's charts and the per-service Trends charts stayed stuck on their rate form. Companion frontend PR: muchq/muchq.github.io#258.
- `request_rate` (unchanged: `sum(rate(...[5m]))`) now sits beside a new, always-present `request_count` (`sum(increase(...[step]))`, bucketed by the chart's own step). Same pairing for `error_rate_percent`/`error_count` — the failure counter's own count/rate pair, not "the error rate as a count".
- No `?view=` on the timeseries endpoint at all, and no key's meaning depends on a query parameter. That's deliberate: threading `?view=` through here would repeat the exact hazard #1287 avoided for the scalar block's own request fields (`RequestsTotal` beside `RatePerSec`, not one field whose meaning flips) — the UI and this proxy deploy independently, so a shared key's meaning changing between deploys is a real mislabeling window regardless of which side ships first.
- `CustomTimeseries` entries are now `customTimeseriesDef` (`tsFixed` for one form, `tsCounter` for a toggleable one) instead of bare query strings. A toggleable entry's map key is a base name that `panels()` expands into `<base>_rate`/`<base>_count` — mirroring how `customScalarDef` already builds two forms from one selector, but as two keys rather than one.
- `microgpt-serve`'s `tokens_per_second` baked one form into its name the way `request_rate` used to, so it's renamed to `tokens` (→ `tokens_rate`/`tokens_count`).
- Added a registry audit (`TestRegistry_CustomTimeseriesPanelKeysAreUniquePerService`) guarding against two `CustomTimeseries` entries silently colliding on the same expanded panel key.

## Deploy order

Serving-chart lookups are safe in either order: an old UI never asks for `request_count`/`error_count` and an old proxy simply has nothing for a new UI's `findSeries('*_count')` to find — no mislabeling either way. Trends is asymmetric: an old UI enumerates every series outside its own hardcoded standard set, so a backend-first rollout would dump `request_count`, `error_count`, and every custom `_count` panel into Trends as unwanted extra charts until the UI catches up. **Frontend-first (or together) avoids that window** — frontend-first against an old backend is safe, since `groupTrends` only pairs a `_rate`/`_count` chart when both siblings exist.

## Test plan

- [x] `go test ./domains/platform/apis/prom_proxy/...` — passing

### DIFF
--- a/domains/platform/apis/prom_proxy/models.go
+++ b/domains/platform/apis/prom_proxy/models.go
@@ -26,15 +26,11 @@ type DataPoint struct {
 
 // TimeSeriesResponse wraps timeseries data with metadata
 type TimeSeriesResponse struct {
-	TimeRange string    `json:"time_range"`
-	StartTime time.Time `json:"start_time"`
-	EndTime   time.Time `json:"end_time"`
-	Step      string    `json:"step"`
-	// Which form request_rate is in (#1287, extended to the Serving charts).
-	// Omitted from the host and container timeseries endpoints, which have no
-	// toggleable series and so never set it — see GetServiceMetricsTimeSeries.
-	View   string       `json:"view,omitempty"`
-	Series []TimeSeries `json:"series"`
+	TimeRange string       `json:"time_range"`
+	StartTime time.Time    `json:"start_time"`
+	EndTime   time.Time    `json:"end_time"`
+	Step      string       `json:"step"`
+	Series    []TimeSeries `json:"series"`
 }
 
 // GetTimeRangeConfig returns duration and step for a given time range
--- a/domains/platform/apis/prom_proxy/models.go
+++ b/domains/platform/apis/prom_proxy/models.go
@@ -26,11 +26,15 @@ type DataPoint struct {
 
 // TimeSeriesResponse wraps timeseries data with metadata
 type TimeSeriesResponse struct {
-	TimeRange string       `json:"time_range"`
-	StartTime time.Time    `json:"start_time"`
-	EndTime   time.Time    `json:"end_time"`
-	Step      string       `json:"step"`
-	Series    []TimeSeries `json:"series"`
+	TimeRange string    `json:"time_range"`
+	StartTime time.Time `json:"start_time"`
+	EndTime   time.Time `json:"end_time"`
+	Step      string    `json:"step"`
+	// Which form request_rate is in (#1287, extended to the Serving charts).
+	// Omitted from the host and container timeseries endpoints, which have no
+	// toggleable series and so never set it — see GetServiceMetricsTimeSeries.
+	View   string       `json:"view,omitempty"`
+	Series []TimeSeries `json:"series"`
 }
 
 // GetTimeRangeConfig returns duration and step for a given time range
--- a/domains/platform/apis/prom_proxy/registry.go
+++ b/domains/platform/apis/prom_proxy/registry.go
@@ -182,9 +182,12 @@ func tsFixed(query string) customTimeseriesDef {
 }
 
 // expandCustomTimeseries flattens a service's Trends descriptors into the
-// panel keys a timeseries response actually carries.
+// panel keys a timeseries response actually carries. A toggleable entry
+// contributes two, so the capacity hint doubles the map's own length —
+// only ever a hint, but len(custom) undercounts by up to 2x whenever the
+// registry is mostly toggleable entries, which is the common case.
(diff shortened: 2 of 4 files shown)


aaylward:
Agreed on the design point — adopted. Switched to the dual-series shape you outlined:

- `request_rate` is back to being byte-identical to before this PR: `sum(rate(http_server_requests_total{...}[5m]))`, unconditionally.
- `request_count` is a new, separate, always-present series: `sum(increase(http_server_requests_total{...}[<step>]))`, bucketed by the range's own step.
- `GetServiceMetricsTimeSeries` no longer reads `?view=` at all — removed the parsing, the 400 path, and `TimeSeriesResponse.View`.

This directly fixes the deploy-order fragility both reviews converged on from opposite sides: your point that a shared-key toggle is a hazard regardless of which side ships first, and the companion frontend review's finding that it *had already happened* — the UI was keying labels off local toggle state while this handler ignored `?view=` entirely, so an old proxy's `rate()` data could get captioned "Requests" as soon as the UI deployed first. With two independent series, an old UI never asks for `request_count` and keeps reading `request_rate` unchanged; a new UI against an old proxy just finds no `request_count` series and shows its ordinary empty state — no mislabeling possible on either side of the rollout. Companion frontend PR (muchq/muchq.github.io#258) now does exactly that: no `?view=` on the timeseries fetch, picks between the two series client-side.

Test nits are moot now that the shape changed — no view param to pin a rate/default case for. Replaced the two `ViewCount`/`ViewRate` pins with query-literal pins for both series, a step-bucketing check, and a test that `?view=` on this route is simply ignored (both series always come back). `go test ./domains/platform/apis/prom_proxy/...` passing.

---

aaylward:
Extended this to cover Error Rate and the per-service Trends charts too (companion frontend PR muchq/muchq.github.io#258 has the UI side).

- `error_count` joins `error_rate_percent` as its own always-present series — the count of failed requests per bucket, not "the error rate as a count" (`error_rate_percent` is already a ratio of two rates and has no window of its own to swap). Same dual-series shape as `request_rate`/`request_count`.
- `CustomTimeseries` entries are now `customTimeseriesDef` instead of bare query strings: `tsFixed` for a chart with one form, `tsCounter` for a toggleable one whose map key expands to `<base>_rate`/`<base>_count` panels — mirroring `customScalarDef`, but as two keys rather than one whose meaning depends on `?view=`, for the deploy-safety reason this thread settled on.
- Fallout: `microgpt-serve`'s `tokens_per_second` baked one form into its name the way `request_rate` used to, so it's renamed to `tokens` (→ `tokens_rate`/`tokens_count`). Grepped to confirm nothing else references the old key.

`go test ./domains/platform/apis/prom_proxy/...` passing.

---

aaylward:
Thanks for the re-review. Response by point:

**1. Deploy order — agreed, called out explicitly rather than left implicit.** You're right that this is asymmetric, not a code defect: Serving-chart lookups are safe either way (an old UI never asks for `request_count`/`error_count`; a new UI against an old proxy just finds nothing to plot), but Trends' catch-all classification means a backend-first rollout would dump every new `_count` panel into Trends as unwanted extra charts on an old frontend. Added a "Deploy order" section to the PR description spelling this out, with the recommendation to ship frontend first or together — frontend-first against an old backend is safe since `groupTrends` only pairs when both siblings exist.

**2. Panel-key uniqueness — fixed.** Added `TestRegistry_CustomTimeseriesPanelKeysAreUniquePerService`: it sums each def's own panel count and compares against the expanded map's size, so a collision (which `expandCustomTimeseries`'s last-wins assignment would otherwise hide) shows up as a count mismatch rather than needing a keys() diff that can't distinguish "two defs, same key" from "one key, as expected."

**3. Nits — all taken:**
- `session_starts` now has a one-line comment explaining why it stayed `tsFixed` rather than becoming a toggleable `session` base key (would've renamed its one client-facing name).
- PR title/body rewritten to describe the dual-series design instead of the abandoned `?view=` framing.
- `expandCustomTimeseries`'s capacity hint now accounts for toggleable entries doubling (`len(custom)*2`).
- `error_count` got the same handler-level step-bucketing pin `request_count` already had (`TestMetricsHandler_GetServiceMetricsTimeSeries_ErrorCountBucketsByTheRangesStep`).

`go test ./domains/platform/apis/prom_proxy/...` passing.

---

PR #1297, opened 2026-08-03
golf_hub: boot-grace reap for restored members (#1295)

### LINKED ISSUE #1295: golf_hub: restored members are never reaped — a deploy turns the 5-minute grace into forever-membership, breaking room share links (muchq.github.io#260)
Split out of the #1276 follow-up work (review-panel advisory on the `c1e3c59` fix) because it turns out to have a live user-facing symptom: [muchq.github.io#260](https://github.com/muchq/muchq.github.io/issues/260), "repeated 'room unavailable or already joined' on room share links."

## The gap

`OnExpired` is the only path that removes a parked member, and it fires from the registry's grace timer — which only exists for sessions admitted on the live instance (`Play` → `registry_.ResumeOrAdd`). `RestoreFromStore` rebuilds room membership from rows but registers nothing with the registry, so a member who was parked (or connected) when the process died is restored with **no grace timer and no reaper**. Their membership persists until they resume and explicitly leave, or the room empties.

Production runs one instance (`deploy/consolidated/compose.yaml`) with `grace_period = 5min` (`golf_hub_main.cc:113`), and every deploy restarts the container. So:

- Live hub, tab closed: membership survives 5 minutes, then reaps. As designed (ADR-0020).
- Hub deployed while parked: membership survives **forever**.

## How this becomes #260

The v2 web client always resumes with its localStorage token, so a returning visitor keeps their player id — and the hub still maps them into their old room. `JoinRoom` refuses anyone already in a room (`hub_handler.cc:711-734`, `"room unavailable or already in a room"`), and the frontend's permalink flow neither leaves the old room first nor clears its join attempt on `commandRejected` — it times out after 10s and retries, forever (`useGolfGame.ts:731-746`). Within the 5-minute grace window that reads as "slow initial sync" (the retry eventually lands); after a deploy it reads as "repeated room unavailable" (it never lands).

The frontend half (leave-before-join on permalinks, clear the attempt on rejection) belongs in muchq.github.io. This issue is the hub half.

## Candidate fixes, not yet decided

1. **Boot-time grace:** on restore, arm a grace-period timer per restored member (a synthetic park), so a member who never resumes within `grace_period` of boot is reaped through the existing `OnExpired` path — including its cross-instance resumed-elsewhere row check, which already guards against reaping a member live on a sibling instance.
2. **Join-time repair:** let `JoinRoom` for a player whose current room membership is not backed by a live registry session implicitly leave the old room first. Smaller blast radius, but it papers over the ghost rather than removing it, and roster views keep showing the ghost until someone joins.
3. Both — (1) is the invariant fix, (2) is UX insurance either way.

Note the presence wrinkle either fix must respect: after `c1e3c59`, restored members read `connected` from their rows (fleet truth). A boot-grace reaper must re-read rows at expiry exactly like `OnExpired` already does, so a member who resumed on a sibling instance mid-grace is not reaped.

## Test shape when picked up

Memory-suite (`hub_e2e_test`), no pg gate: park a member in room X, rebuild the hub over the same store (`BuildSecondInstance`), advance/shrink grace, and assert (a) the ghost is reaped without any resume, (b) a resumed player parked in X can follow a share-link join into room Y — named after the #260 symptom.

aaylward:
## Summary

The hub half of muchq.github.io#260 (frontend half: muchq.github.io#266). The registry's ADR-0020 grace only covers seats it admitted, so a restart converted every parked member's five-minute grace into forever-membership: the successor restored them from rows, and nothing ever reaped them. That fossil membership is what answered share-link joins with `"room unavailable or already in a room"` until the visitor's old room emptied by other means.

## Fix

`RestoreFromStore` records a boot cohort and arms one deadline of the same `grace_period`; whatever no session reclaims by then reaps through the same fresh-row check as session expiry (`OnExpired` → `ReapUnlessResumedElsewhere`, factored). Timing only ever favors the player: park at T, deploy at D, reap at boot+5min ≥ T+5min.

The cohort's contract: **a seat whose row reads connected belongs to whatever session owns that row.** It never enters the cohort at restore, and is ceded when any reconcile observes the row connected. The cede is a sample, not a subscription (a pulse no wake shows this instance still early-reaps in a fleet) — owned in a comment and tracked with the rest of the presence-ownership residue in #1299. The drain re-checks stop between reaps so `~HubHandler` waits out one store round trip, not the cohort; `RestoreFromStore` refuses a second call via a real `restored_` flag (not `joinable()`, which a thread-less first restore would fool). The reaper reports through `restored_seats_reaped` (prom_proxy Sessions tile beside `seats_expired` — it fires unattended at boot+grace, so the series is the only evidence it runs at all).

**Deliberately not covered** (recorded on #1295 and tracked in #1299): hard-crash ghosts whose rows still read connected are spared — indistinguishable from a sibling's live seat without row-level presence ownership. Orderly restarts (deploys — the #260 case) write the goodbyes and are fully covered.

## Tests

All memory-suite — no pg gate, so they run on every CI pass. Each guard has its own pin; each reap test carries a parked-ghost canary hard-asserted gone, so a reaper that never fires cannot pass vacuously:

| Test | Pins |
|---|---|
| `BootGraceReapsParkedGhostAndSparesLiveSeats` | happy-path smoke of the whole flow (explicitly not a unique pin of any one guard) |
| `ShareLinkJoinSucceedsOnceBootGraceClearsStaleMembership` | both #260 timelines: mid-window resume + bare join still gets the exact refusal (the frontend half's territory), and the post-reap join lands |
| `BootGraceNeverClaimsASeatRestoredConnected` | restore-time cohort filter (awaits the parked row before watching the reap) |
| `BootGraceCedesASeatObservedConnectedMidWindow` | reconcile-time cede |
| `BootGraceDrainSparesARowThatTurnedConnectedUnobserved` | the drain's fresh-read row check — the one shape neither filter nor cede can reach |
| `RestoreFromStoreRefusesASecondCall` | the once-guard, including the thread-less first restore |

`BuildSecondInstance` uses the shared `RemoteIdGenerator` (now one class in `id_generator.h` beside `SequentialIdGenerator`; the altitude pass caught this PR introducing its fourth byte-identical copy) — a fresh seat on the second instance used to mint the first instance's `player-1` and refuse that player's own resume as a seat conflict.

| Check | Result |
|---|---|
| `hub_e2e_test` (11.9s, `size=small`), incl. `--runs_per_test=15` | PASSED |
| Full golf_hub + pg suites vs postgres 16 | 16/16 PASSED |
| `scripts/mutation-check`: filter unconditional, cede deleted, drain row check forced, once-guard disabled | all 4 killed |
| clang-format | clean |

Review history: the panel round (concurrency + semantics lenses; the test-quality agent was stopped mid-run after its in-tree mutation runs collided with live edits — its checks were re-run from the main loop) surfaced the fleet early-reap, fixed from both ends. The follow-up cursor review's should-fixes are all in, and the altitude pass (#1298's fourth lens) confirmed the level — its two survivors (the invisible reaper metric, the fourth `RemoteIdGenerator`) landed in `5c90e37`, with the clock seam filed as #1300. The working-agreement process edit moved to #1298 per "One item, one PR".

Fixes #1295.

https://claude.ai/code/session_01VUrPZEoRVcJ6qUe794xhvD

### DIFF
--- a/domains/games/apis/golf_hub/hub_chat_race_test.cc
+++ b/domains/games/apis/golf_hub/hub_chat_race_test.cc
@@ -113,20 +113,6 @@ class GatedChatStore final : public ChatStore {
   TestGate released_;
 };
 
-// Secondary instances mint from their own space so two hubs can never
-// hand out the same player id.
-class RemoteIdGenerator final : public IdGenerator {
- public:
-  std::string PlayerId() override { return "remote-player-" + std::to_string(++players_); }
-  std::string RoomId() override { return "remote-room-" + std::to_string(++rooms_); }
-  std::string GameCode() override { return "RGAME" + std::to_string(++games_); }
-
- private:
-  int players_ = 0;
-  int rooms_ = 0;
-  int games_ = 0;
-};
-
 class HubChatRaceFixture : public GolfHubStreamFixture {
  protected:
   struct Instance {
--- a/domains/games/apis/golf_hub/hub_e2e_test.cc
+++ b/domains/games/apis/golf_hub/hub_e2e_test.cc
@@ -204,23 +204,6 @@ class NotAMemberChatStore final : public ChatStore {
   std::shared_ptr<ChatStore> delegate_;
 };
 
-// Distinct id space for the second instance (the hub_store_race_test /
-// pg suite pattern): two SequentialIdGenerators would both mint
-// "player-1", and a fresh seat on the second instance would then hold
-// the first instance's player id — a seat conflict for that player's
-// own resume.
-class RemoteIdGenerator final : public IdGenerator {
- public:
-  std::string PlayerId() override { return "remote-player-" + std::to_string(++players_); }
-  std::string RoomId() override { return "remote-room-" + std::to_string(++rooms_); }
-  std::string GameCode() override { return "RGAME" + std::to_string(++games_); }
-
- private:
-  int players_ = 0;
-  int rooms_ = 0;
-  int games_ = 0;
-};
-
 std::unique_ptr<SecondInstance> BuildSecondInstance(
     std::shared_ptr<TicketVault> vault, std::shared_ptr<HubStore> store,
     std::shared_ptr<ChatStore> chat_store,
--- a/domains/games/apis/golf_hub/hub_e2e_test.cc
+++ b/domains/games/apis/golf_hub/hub_e2e_test.cc
@@ -1103,6 +1103,132 @@ TEST_F(GolfHubStreamFixture, ShareLinkJoinSucceedsOnceBootGraceClearsStaleMember
   EXPECT_EQ(joined->as_roomState_or_null()->players.size(), 2u);
 }
 
+// The fleet half of the boot cohort's contract (#1295): a seat restored
+// with its row at connected belongs to whatever session owns that row —
+// here, alice live on this generation — and must never enter the
+// sibling's cohort at all. Otherwise her park during the sibling's boot
+// window would be reaped at its boot deadline instead of running her
+// own full grace on this generation's registry. bob, parked before the
+// boot with his row at disconnected, is the canary that proves the
+// deadline fired and the drain ran.
+TEST_F(GolfHubStreamFixture, BootGraceNeverClaimsASeatRestoredConnected) {
+  auto alice = OpenSeat();
+  auto bob = OpenSeat();
+  ASSERT_TRUE(alice.has_value() && bob.has_value());
+  ASSERT_TRUE(ReceiveCase(alice->stream, "sessionReady").has_value());
+  ASSERT_TRUE(ReceiveCase(bob->stream, "sessionReady").has_value());
+  const std::string room_id = CreateRoomFor(*alice);
+  ASSERT_FALSE(room_id.empty());
+  moonbase::golf::JoinRoom join;
+  join.roomId = room_id;
+  ASSERT_TRUE(bob->stream.Send(GolfCommands::FromJoinroom(join)).ok());
+  ASSERT_TRUE(ReceiveCase(bob->stream, "roomState").has_value());
+  bob->stream.Close();
+  AwaitSnapshot(*store_, [&](const HubStore::Snapshot& snapshot) {
+    for (const auto& member : snapshot.members) {
+      if (member.player_id == bob->player_id) return !member.connected;
+    }
+    return false;
+  });
+
+  // The sibling boots while alice is live: her row reads connected.
+  auto instance =
+      BuildSecondInstance(vault_, store_, chat_store_,
+                          std::make_shared<futility::otel::MetricsRecorder>("golf_hub_test"),
+                          /*grace=*/std::chrono::seconds(1));
+  ASSERT_NE(instance, nullptr);
+  // She parks inside the sibling's boot window; this generation's
+  // registry arms her grace (60s here — far past the test).
+  alice->stream.Close();
+
+  // bob's reap is the deadline's proof; alice must outlive it.
+  auto reaped = AwaitSnapshot(
+      *store_,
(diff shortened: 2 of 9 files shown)


aaylward:
All four should-fixes are in (`be2e7af`), plus the advisories that warranted action. Thread-by-thread:

- **ShareLink overclaim** → took option (b): the test now pins both timelines. dave resumes mid-window into the live fossil and his bare join receives the exact `"room unavailable or already in a room"` — labeled as the frontend half's territory (muchq.github.io#266 ships the leave-before-join) — and alice's post-reap resume + join still lands. The comment and PR table no longer claim more than the hub owns.
- **Drain row check unpinned** → the sharpest catch here: the restore filter had indeed orphaned that mutant (`resumed_elsewhere = false` survived nothing structurally after the filter kept connected rows out of the cohort). New `BootGraceDrainSparesARowThatTurnedConnectedUnobserved` reaches the drain check in the one shape neither filter nor cede can: a cohort member turns connected via a resume on the sibling generation with no reconcile ever showing this instance the pulse. Mutant re-killed.
- **Soft canary in Cedes** → hard `ASSERT_FALSE(survivors.contains(carol))` before the spare assertions; same hardening applied to the new tests.
- **Second-restore refusal untested / `joinable()` sentinel** → both: a real `restored_` flag (your advisory's shape — a thread-less first restore must refuse all the same) and `RestoreFromStoreRefusesASecondCall` covering the armed and thread-less cases. The `if (restored_)` guard is mutation-checked.
- **NeverClaims timing** → awaits alice's parked row before watching for the canary reap, per your suggested remedy; acknowledged in-comment that the remaining bet is wall-clock, not structural.
- **Connected-pulse miss (fleet)** → owned in the cede comment ("a sample, not a subscription") and tracked with the rest of the presence-ownership residue in #1299, which also answers the residue-ownership question now that this PR closes #1295.
- **WORKING_AGREEMENT edit in the product PR** → agreed; moved to #1298.

Full battery after the changes: `hub_e2e_test` 11.8s green, golf_hub + pg suites 16/16 vs postgres 16, all four guard mutants killed.

---

aaylward:
Ran the altitude lens (#1298's fourth seat) over this PR post-approval — read-only agent, findings verified in the main loop. Verdict: **the shipped shape is right at the level it sits**, with two survivors, both landed in `5c90e37`:

- **The reaper was invisible in production.** `restored_seats_reaped` fires unattended at boot+grace and nothing reported on it — a reaper that silently stops arming is indistinguishable from a healthy hub with no ghosts, which is exactly the "zero that means healthy and also broken" shape the testing bar warns about. The series now sits in prom_proxy's golf_hub catalog beside `stream_seats_expired_total` (positional catalog tests shifted along).
- **`RemoteIdGenerator` had hit four byte-identical copies** (this PR added the fourth). It now lives beside `SequentialIdGenerator` in `id_generator.h` — the test-only-generator precedent that header already set — net −37 lines, no new dep edges.

The level checks it refuted, for the record: moving absent-seat grace into smithy's `SessionRegistry` fails concretely (`Drain` fires `on_expired` for every detached entry on SIGTERM, so parked-restored members would be mass-reaped on each deploy — the opposite of the fix — and ADR-0020 explicitly scopes restart survival out as "application persistence"); deriving the reap set from mirrored room state instead of the cohort fails because `ReconcileRoomLocked` copies the row's `connected` into memory, and a level cannot answer whose registry owns a seat's timer — that takes the edges observed during the window, which is what the cohort records; and the frontend's leave-before-join alone cannot reclaim a room made entirely of ghosts (nobody joins one), so the reaper is the only path to `DeleteRoom`'s cascade for fossil rooms.

Also confirmed the item the altitude agent could only flag: `hub_e2e_test` survives `--runs_per_test=15`, and all four guard mutants were re-verified killed from the main loop earlier. Follow-up filed for the missing clock seam shared with the registry (#1300).

---

Issue #1305, opened 2026-08-04
futility: http_server_* route label is the raw request path — unbounded series, and a vocabulary the other rails don't speak


aaylward:
Surfaced by #1303's altitude review. futility has emitted a `route` label since #1174, but its value is the raw request path: `aura/middleware.cc`'s `RouteOf(target)` is `target.substr(0, target.find('?'))`. Two consequences:

1. **Unbounded cardinality, live today.** A scanner (or any path-shaped ID scheme) against golf_hub or portrait mints a Prometheus series per distinct path. This is exactly the shape #1303 refused to ship in yodel — its route is the matched template with an `unmatched` sentinel, and its tests pin the boundedness. futility should bound the same way: matched handler route where the router knows it, sentinel otherwise.

2. **The rails share a label name but not a vocabulary.** yodel: matched template. futility: raw path. server_pal: no route at all (#1304). Literal-route queries (`route="/health"`, which prom_proxy's probeFilter relies on) work on every rail because a literal path and its template coincide; anything template-shaped, or relying on bounded cardinality, is yodel-only. Related timing divergence, also unpinned: yodel's `requests` counts at completion (#1303, so it can carry the route); futility and server_pal count at start. Nothing queries `requests - success - failure` today, and error rate is computed as failure/(success+failure), so nothing breaks — but `otel_contract` pins descriptions and bucket layouts only, so nothing would notice these drifting further apart either. When this is picked up, add a label-set (and, in prose, timing) row to the contract alongside the description pins; yodel's README now carries the honest dialect table in the meantime.

Issue #1304, opened 2026-08-04
server_pal: http_server_* carries no route label — the yodel gap of #1303, one HTTP probe away from mattering


aaylward:
Split from #1303, which fixed the same gap in yodel. `server_pal`'s middleware labels the `http_server_*` family with only `http_method` and `service_name` (`lib.rs`, `http_metrics_middleware`), so probe traffic on any server_pal-backed service would be inseparable from serving traffic — exactly the one_d4 symptom #1303 diagnosed.

Latent today: no server_pal service in `deploy/consolidated/compose.yaml` has an HTTP-probing healthcheck (they use `pg_isready` and friends), so nothing is currently polluted. prom_proxy's Serving expressions already exclude `route!="/health"`, and a negative matcher matches label-less series, so server_pal services lose nothing now and start benefiting the moment the label exists.

When picked up: copy the route **concept**, not the spelling divergence — the label is `route` (as in futility and now yodel), keyed on the matched route template rather than the raw path so the series stay bounded, with a fixed sentinel for unmatched requests (yodel uses `unmatched`). Note yodel's completion-time recording contract (#1303): the route is only knowable after routing, so the counters move at completion and only the in-flight gauge moves at start, without a route. Axum's `MatchedPath` extractor is the likely source of the template. Guard shape as in #1303: a test that the request counter carries the route, and that the gauge does not.

Issue #1174, opened 2026-07-17
[portrait] Post-soak teardown: retire the meerkat serving path


aaylward:
Follow-up to #1168 (complete through Phase 4: #1169, #1170, #1171, #1172, #1173). `ghcr.io/muchq/portrait` ships the smithy-cpp server as of `main@40d3e59`. This issue tracks the final teardown once the new binary has soaked; the working list also lives in `domains/graphics/apis/portrait/PORTRAIT_TODO.md`.

## Soak gate (do not start before)

One release (~a week) in production on the smithy binary with:
- [ ] `http_server_*` dashboards unchanged (instrument parity was by construction — verify in practice)
- [ ] no 5xx / `error_type` regressions; cache-hit latencies sane at microsecond granularity
- [ ] log correlation via `trace_id=` intact for whatever consumes the access logs
- [ ] no rollback to the pre-cutover image tag needed

## Teardown checklist

1. **Delete the meerkat serving path from portrait**
   - `:portrait_meerkat` + `Main.cc`
   - `portrait_parity_test` (+ BUILD target) — its differential purpose ends with the meerkat stack
   - The nlohmann wire serde in `types.h` (`NLOHMANN_DEFINE_TYPE_*` macros, enum serialization) and the `@nlohmann_json` dep on `:types` — nothing parses JSON into `portrait::TraceRequest` after the above are gone. **Keep** the plain structs, `validate*` cross-field rules, and the absl hashes: they are `TracerService`'s domain layer and the cache key. Trim `types_test.cc` to validation coverage.
2. **`TracerService` hands back raw PNG bytes** — return raw bytes + dimensions instead of a base64 string; cache the raw bytes; `smithy_handler.cc` wraps them in the `Blob` directly. Deletes the base64 encode→decode→re-encode round-trip, `imageToBase64`, and portrait's `futility/base64` dep. Update `tracer_service_test`.
3. **Test consolidation** — a small `testonly` library with the canonical valid-scene fixture (JSON + typed) and the loopback harness for the three remaining test files (`portrait_smithy_wire_test`, `smithy_handler_test`, `smithy_middleware_test`); extract the production middleware chain from `portrait_smithy_main.cc` into a shared builder so the middleware tests exercise the real wiring instead of a reconstructed copy.
[...]

PR #1306, opened 2026-08-04
yodel: route label on http_server_* so probes are separable from serving traffic

### LINKED ISSUE #1303: yodel: http_server_* has no route label, so healthcheck probes are indistinguishable from serving traffic
The one_d4 Serving charts show a flat ~10 requests / 5 min for the whole day, which reads as sustained traffic. All of it is the container's own healthcheck.

`deploy/consolidated/compose.yaml` probes `GET /health` on `interval: 30s`, which is 10 requests per 5-minute bucket. Live `request_count` over 1d confirms it: 195 of 199 non-empty buckets are exactly 9.474 or 10.526 — 9 or 10 raw probes, scaled by the 20/19 extrapolation `increase()` applies to a 5m window over 30s scrapes. Only four buckets in 24h deviate (4.2, 8.4, 13.7, 21.1); those are the only real requests in the window.

The series steps off zero at `2026-08-03T20:27:02Z`, which is when #1293 deployed. Before that the probe shelled out to `curl`, absent from `@java_base`, so it exited on "executable file not found" and issued no HTTP request at all. Fixing the probe is what made the floor appear — the probe is correct now, and the traffic it generates is genuine.

Nothing downstream is wrong. `registry.go:500` is `sum(increase(http_server_requests_total{service_name="one_d4"}[5m]))` at step `5m` (`models.go:41-42`) — non-overlapping buckets, no double counting — and the dashboard plots what it is handed without accumulating.

The problem is that the probe cannot be separated from real traffic. yodel's `http_server_*` family carries only `http_method` and `service_name` (`OtlpJsonEncoder.java:246-253`, over `HttpServerMetrics.MethodSnapshot` at `HttpServerMetrics.java:98-106`). There is no route dimension, so no PromQL selector can exclude `/health`, and prom_proxy and the dashboard both inherit a floor they have no way to subtract. Every one_d4 serving number carries it: REQ/S sits at ~0.035 on an idle service, REQ (5M) never drops below ~10, and the Requests chart never returns to zero.

futility already emits the missing dimension — `http_metrics.cc:48` returns `{{"service_name", ...}, {"route", route}, {"method", method}}` — so the shape of the fix exists in the repo. Note it spells the method label `method` where yodel and server_pal use `http_method`; copy the route concept, not the spelling, or the families diverge.

Suggested fix: add a route (or a boolean probe) attribute to yodel's http_server_* family, keyed on the matched route rather than the raw path so the label stays bounded, and have prom_proxy's `standardTimeseriesQueries` exclude probe traffic from the Serving tiles. Whether `/health` should be excluded outright or merely separable is the open call — it is real load and worth being able to see, just not worth being the only thing the Serving section shows.

Guard: a test asserting the request counter carries a route attribute, and a prom_proxy query test asserting the serving expressions exclude the probe route.

Scope: one_d4 is the only service in `deploy/consolidated/compose.yaml` with an HTTP-probing healthcheck, so it is the only one showing this today. The other services are clean because they have no HTTP probe (most use `pg_isready`), not because their instrumentation distinguishes probe traffic — server_pal has the same gap as yodel (`lib.rs:216-219` sets only `http_method` and `service_name`), so any yodel- or server_pal-backed service that adopts an HTTP probe inherits this. futility is the only rail already immune.

aaylward:
Closes #1303.

## What

Every yodel `http_server_*` counter and histogram point now carries a `route` label — the **matched route template** (`/widgets/{id}`), never the raw path — and prom_proxy's standard Serving queries exclude `route="/health"`, so container-probe traffic stops inflating request counts and diluting error rates. one_d4's dashboard gets a **Probes** tile showing the excluded traffic, so the subtraction is a visible fact rather than a silent floor under every chart.

## Why now

This floor is two days old. Before #1293 the one_d4 healthcheck never actually ran (the image had no `curl`), so `/health` traffic never existed. #1293 fixed the probe, and every 30s it has been feeding the Serving numbers since — ~10 requests/5m of permanent baseline plus a steady stream of fast 200s dragging down avg/p95 latency.

## Design decisions

- **Route label, not a probe boolean.** Tagging probe requests `probe="true"` at the filter was the alternative. Rejected on three grounds: *policy location* — what counts as a probe is a dashboard-side decision, and hardcoding it into each service's filter means redeploying every emitter when the policy changes, where `probeFilter` is one Go constant; *generality* — a route label also answers per-route questions we haven't asked yet, a probe boolean answers exactly one; *parity* — futility already spells the label `route`, so this converges the rails on a name instead of minting a new one.
- **Matched template, never the raw path, with an `unmatched` sentinel.** A scanner spraying paths (or invented HTTP verbs — method is bounded via the enum for the same reason) cannot mint unbounded series.
- **Counters move to completion; the gauge stays method-keyed.** The filter deliberately runs before routing so unmatched requests are measured, which means the route is only knowable at completion. Once in-flight requests drain, `requests == success + failure + abandoned`; the snapshot reads outcomes before requests (the writer's reverse), so a torn read can never show outcomes running past requests.
- **The negative matcher is fleet-safe, gauge included.** `route!="/health"` matches series that lack the label entirely, so applying it to every standard selector is a no-op for server_pal (no route label, #1304) and correct subtraction for futility (raw-path label since #1174 — bounded-cardinality is yodel-only for now, #1305).

## Guards

The `/health` literal is pinned at all three of its homes, so no single-sided edit can silently re-inflate the numbers:

| Guard | Pins |
|---|---|
| `HttpServerMetricsFilterTest.routesAreTheMatchedTemplatesAndUnmatchedIsASentinel` | Template values (via a parameterized route, the one shape where template ≠ raw path), sentinel, bounded series count |
| `OtlpJsonEncoderTest` route tests | `route` attribute on counters + histogram, absent on the gauge |
| `TestRegistry_StandardServingQueriesExcludeTheProbeRoute` | Every `http_server_*` selector in both query sets carries `route!="/health"` |
| `HealthProbeRouteLabelTest` (one_d4) | Probe requests arrive under exactly `/health` across the micronaut-jaxrs bridge — the leg nothing else pins |
| `TestOneD4HealthcheckProbesTheRouteProbeFilterSubtracts` (deploy) | The compose healthcheck actually probes `GET /health` |
| `aSynchronouslyThrowingChainStillCountsTheRequest` | A chain that throws before producing a publisher still lands in the counters |

Mutation battery (main loop, clean tree): **13/13 killed** — routeOf falling back to the raw path (both shapes), `Mono.defer` removed, abandon losing its request count, start gaining one, route dropped from the counter and histogram encoders separately, route added to the gauge, `probeFilter` emptied, dropped from a scalar selector / a timeseries selector / the gauge selector individually, and the Probes tile pointed at `/healthz`.

## Review

Four read-only panel lenses ran (correctness/control-flow, data-access/resource-safety, tests/docs/CI, altitude), each refuting its own findings; all surviving findings are addressed in the diff. Notable: the data lens claimed the route attribute is never populated (static jar scan) — refuted empirically by the two integration tests, which observe real templates stamped on native and JAX-RS routes. Correctness findings landed as: `BasicHttpAttributes.getUriTemplate` (the supported accessor; the raw constant is deprecated forRemoval), `Mono.defer` around `chain.proceed`, and the torn-snapshot read ordering.

Follow-ups filed rather than smuggled in: #1304 (server_pal route label), #1305 (futility's raw-path values).

https://claude.ai/code/session_01VUrPZEoRVcJ6qUe794xhvD

---
_Generated by [Claude Code](https://claude.ai/code/session_01VUrPZEoRVcJ6qUe794xhvD)_


aaylward:
Both gaps closed in 8d68d1f, each with its mutation verified:

1. **Probes tile existence** — `TestOneD4KeepsTheProbesTile` asserts some one_d4 custom scalar query contains `route="/health"`. Deleting the tile line now fails the suite (checked with a line-deletion mutation; the wrong-path shape was already killed by the carve-out).
2. **Method-enum bounding** — `aCustomVerbCollapsesToTheBoundedMethodEnum` sends a real `BREW` request over the socket and asserts both the route snapshots and the gauge carry only `CUSTOM`. The one-token revert to `getMethodName()` now fails the suite (mutation-checked).

---

Issue #1302, opened 2026-08-04
chessql: NULL titles silently drop untitled opponents from every negated title filter


aaylward:
Asking for a player's games against non-GM opponents silently drops every untitled opponent — usually most of the intended set.

`opponent.title != "GM"` and `NOT opponent.title = "GM"` both compile to a bare inequality (`SqlCompiler.java:447-450`):

```sql
LOWER(CASE WHEN LOWER(white_username) = LOWER(?) THEN black_title ELSE white_title END) != LOWER(?)
```

`white_title`/`black_title` are nullable with no default (`Migration.java:41`, `:98`, `:407`), and untitled players are written as NULL: `fetchTitle` maps an absent profile title to `null` (`IndexWorker.java:1011-1019`), `titleFor` returns null for it (`IndexWorker.java:1025-1030`), and the DAO binds it straight through (`GameFeatureDao.java:180`).

So for an untitled opponent the predicate evaluates to NULL, not TRUE, and the row is excluded from the WHERE clause. The caller gets a smaller number with no error and no indication anything was dropped.

Same defect on the other operators over the same columns:
- `white.title != "GM"` / `NOT black.title = "GM"` — the non-perspective path, `SqlCompiler.java:455-457`
- `NOT opponent.title IN ["GM", "IM"]` — `SqlCompiler.java:550-561`; SQL `NOT IN` is NULL-yielding for the same reason

Detection is hard from the outside: the result is a plausible-looking count. It only surfaces if you separately aggregate by title and notice the totals don't reconcile. I hit this on `oleksandr_bortnyk` and it happened not to bite — all 32 of his bullet opponents that month were titled, so `NOT opponent.title = "GM"` returned a correct 23 — but that is luck, not correctness. For a typical non-GM account whose opponents are mostly untitled, the same query returns near-zero.

There is a second conflation underneath: `titleFor` also yields null when the profile fetch *failed* (`TitleResolution.degraded`, `IndexWorker.java:966-993`), so NULL means both "untitled" and "unknown". Any fix should decide which of those a negated title filter includes.

Suggested fix: make negated string comparisons NULL-safe in the compiler rather than at each call site. For `!=` on a nullable string column, emit `(col IS NULL OR LOWER(col) != LOWER(?))`; for `NOT IN`, the equivalent. `NotExpr` at `SqlCompiler.java:424` wraps an arbitrary subexpression in bare `NOT`, so it needs the same treatment — a NULL inner predicate stays NULL through the negation.

[...]

PR #1311, opened 2026-08-05
chessql: every perspective field is groupable — categorical by value, ratings as buckets

### LINKED ISSUE #1301: chessql: only me.color and outcome are groupable — no way to aggregate by opponent title/username across both colors
`SqlCompiler.java:358-359` allows only two perspective fields in `groupBy`:

```java
private static final Map<String, String> GROUPABLE_PERSPECTIVE_FIELDS =
    Map.of("me.color", "me.color", "me_color", "me.color", "outcome", "outcome");
```

The other five (`me.elo`, `me.title`, `opponent.username`, `opponent.elo`, `opponent.title`) are filter-only and rejected at `SqlCompiler.java:380-384`. They already compile to CASE expressions in `PERSPECTIVE_FIELDS` (`SqlCompiler.java:111-136`), which is the same machinery `me.color` and `outcome` use to reach the SELECT list, so the grouping path is mostly built already.

The gap shows up on any "who did I play, and how did it go" question. `aggregate` with `player: X` and filter `opponent.title != "GM"` works; grouping the same corpus by `opponent.title` does not. The workaround is to group by the color-specific column instead — `white_title` when the player was Black, `black_title` when White. That is only correct if the filter already pins the player to one color. A both-colors aggregate has no correct workaround: `white_title` mixes the player's own title into the opponent's buckets on half the rows, silently.

Concrete: for `oleksandr_bortnyk`, bullet, one month, `me.color = "black"` — grouping by `white_title` gives the opponent title breakdown (GM 9, FM 8, IM 7, CM 6, NM 2). Drop the `me.color` clause to ask the same question across both colors and there is no field that yields it.

Suggested fix: add the remaining five to `GROUPABLE_PERSPECTIVE_FIELDS`, keyed by both dotted and underscore spellings, matching the existing `me_color` convention (`opponent_title`, `opponent_username`, `opponent_elo`, `me_title`, `me_elo`). Numeric fields (`*.elo`) group into one bucket per distinct rating, which is rarely what a caller wants — either leave them out or document that they need bucketing at the call site.

`AggregateGamesTool.java` and the MCP tool descriptions list the groupable fields inline and would need the same update.

Separately, and worth its own issue: `opponent.title != "GM"` / `NOT opponent.title = "GM"` compiles to a bare `LOWER(CASE ...) != LOWER(?)` (`SqlCompiler.java:447-450`), and `white_title`/`black_title` are nullable with no default (`Migration.java:41`, `:98`, `:407`). Untitled opponents are NULL, the predicate is NULL, and the rows drop — so "non-GM opponents" silently excludes exactly the untitled players who make up most of that set. Same shape as #1256.

### LINKED ISSUE #1310: chessql: rating buckets for aggregate groupBy (me.elo / opponent.elo)
Split from #1301, which made every *categorical* perspective field groupable and deliberately left the two rating fields filter-only: GROUP BY on a raw rating makes one bucket per distinct value, which buries the answer under hundreds of one-game groups and then hits the caller's group limit — the silent-truncation shape #1290-era `/v1/aggregate` was already burned by. The rejection message points callers at per-band range filters (one aggregate call per band) in the meantime.

The real feature is bucketed grouping, and it has genuine design choices rather than a mechanical fix, which is why it's its own item:

- **Bucket width**: fixed (e.g. 100 or 200 points), caller-supplied, or relative to the player's own rating (the question behind "how do I do against stronger players" is usually a *delta*, not an absolute band).
- **Band edges and labels**: half-open ranges, what the group key looks like on the wire (`"2400-2499"`? a numeric lower bound?), and an open-ended top band.
- **NULL elos**: rows indexed before the columns existed hold NULL; a NULL band needs the same explicit treatment the title fields' null bucket got in #1301.
- **Portability**: the arithmetic (`FLOOR(elo/width)*width` or a CASE ladder) must compile identically on H2 and Postgres, and the SELECT-alias-in-GROUP-BY convention from #1301 should carry over.

`SqlCompiler`'s `GROUPABLE_PERSPECTIVE_FIELDS` javadoc and the rejection message reference this issue as the tracked follow-up.

aaylward:
Closes #1301. Closes #1310.

`groupBy` on `/v1/aggregate` (and the `aggregate_chess_games` MCP tool) now accepts every perspective field when `player` is supplied, in either spelling, keyed by underscore forms in the response:

- **Categorical fields group by value** (#1301): `me.color`, `me.title`, `opponent.username`, `opponent.title`, `outcome`. Grouping by `opponent.title` / `opponent.username` is the only correct opponent breakdown across both colors — the physical color-specific columns hold the *player's own* value on half the rows, so grouping them mixes the player into the opponent buckets. The issue's motivating query ("hikaru's results against each GM he faced, both colors pooled") works in one call.
- **Rating fields group as fixed-width buckets** (#1310, folded in on request): bare `me.elo` buckets by 100 points; a parenthesized width overrides it (`opponent.elo(200)`, mirroring the existing `motif_count(pin)` syntax). The group key is the band's numeric lower bound, rendered as `(CASE ...) / width * width` on the aliased SELECT expression and inlined in the totals' GROUP BY.

In the compiler, groupability is a property of the field: `PerspectiveField` carries a `GroupKind` (CATEGORICAL / RATING), both accepted spellings derive mechanically from the canonical names, and the error-message rosters derive from the same source — an eighth perspective field is one map entry, and spelling drift is unrepresentable rather than tested for.

### Decisions

- **Absolute bands, caller-supplied width; player-relative (delta) bands not implemented.** #1310 posed "how do I do against stronger players" as possibly a delta question. Only absolute bands ship — a delta group key is a different derived quantity (opponent minus me), negative-valued, with its own floor semantics around zero. If wanted, it's a fresh issue.
- **No open-ended top band.** The top band is just another width-sized band; super-GM tails shrink but can still fragment. The untruncated `totalGroups` / `truncated` machinery is what reports that.
- **Numeric lower bound as the key, not a `"2400-2499"` label.** A rendered label would sort lexicographically in the ASC tiebreak (`"800-899"` after `"1000-1099"`); the numeric key sorts numerically on both engines and reaches JSON as a number (`"opponent_elo": 2400`). API.md shows the numeric and `null`-bucket rows explicitly.
- **Integer division (`expr / width * width`), not `FLOOR(elo/width)*width`.** Not argued from arithmetic: `aggregate_bucketArithmeticAtIntegerExtremes` (H2) and `aggregateBucketArithmeticAtIntegerExtremesOnPostgres` (real dialect, CI) execute the extremes — an `Integer.MAX_VALUE` elo keys 2147483600 at width 100 and itself at width `Integer.MAX_VALUE` (everything else collapsing to bucket 0, no overflow, key still an Integer), and a schema-possible negative elo truncates toward zero (-150 → -100, where FLOOR would give -200) identically on both engines.
- **Width 1 is legal, pinned, and documented** — the rejection guards against *implicit* raw grouping; an explicit `opponent.elo(1)` is a deliberate opt-in that `totalGroups`/`truncated` still report on. Width 0 or less, malformed, overflowing, and unclosed-paren widths are 400s with actionable messages; **conflicting widths for one field are rejected** rather than silently first-one-wins (same-width duplicates dedupe like spelling duplicates).
- **NULL group keys stay NULL** at every level (compiler SQL, H2, real Postgres, controller map, REST wire, tool JSON). For titles that's untitled opponents; for elos it's games where chess.com omitted a side's player-result record — pinned at the source by `process_gameMissingASidesResultIndexesWithNullUsernameEloAndTitle` (the worker persists the side as NULL rather than failing the game or inventing a zero). Docs call out the title null bucket as the practical way to *count* untitled opponents while `!=` excludes NULLs (#1302). **Review caught that the REST wire violated this** — the HTTP mapper's default inclusion dropped null map values, and `group` entirely once all values were suppressed (the MCP path was unaffected; `JsonUtils` serializes with ALWAYS). Fixed by pinning `@JsonInclude(ALWAYS, content = ALWAYS)` on `AggregateRow.group` alone; `IndexResponse`'s null-field omission is untouched.
- **`opponent.username` groups by stored casing** (filter matches case-insensitively; group keys are not normalized) — documented and executed: `aggregate_opponentUsernameGroupsKeepStoredCasingDistinct` proves one opponent stored as `Foe`/`foe` forms two groups while the player match stays case-insensitive in the same fixture.
- **NULLS-LAST tiebreak not added.** The divergence itself is now a pair of CI tests, not a comment: `aggregate_nullGroupKeySortsFirstInTheTiebreakOnH2` and `nullGroupKeySortsLastInTheTiebreakOnPostgres` (real dialect, CI) pin opposite orders on purpose — a compiler normalization or an engine default change fails exactly one twin. Sibling tests with tied counts stay order-free.

### Verification

- `bazel test` over chessql, one_d4, mcpserver: green locally. `scripts/diff-build origin/main` green on the diff scope.
- **`AggregateResponseWireTest`** (new, mirrors `IndexResponseWireTest`): a real server's raw `/v1/aggregate` bytes keep explicit null group keys (`"group":{"opponent_elo":null}`), keep the null dimension beside non-null keys in multi-key groups, and serialize non-null bucket keys as JSON numbers. Both tests failed on exactly the dropped-key bodies before the inclusion fix.
- **`PostgresAggregateCompatTest` compiles locally but executes only in CI** (`PG_TEST_DB_URL`-gated); CI's build-and-test runs it for real (confirmed passing, not skipped, by review). Its four new tests cover the categorical null bucket, the bucket arithmetic + alias + NULL propagation, the INT-extreme arithmetic, and the NULLS-last tiebreak order.
- **A claims audit closed the reasoning-only gaps**: every claim this PR had settled by argument or source-reading — dialect NULL ordering, stored-casing grouping, NULL-elo provenance, INT-extreme arithmetic — now has a test that executes it (on both engines where the claim is about both engines).
- **Mutation checks, 18 killed on clean trees**, including: dropping a groupable map entry / flipping a field's `GroupKind`; flipping each CASE to the wrong side (caught through H2, not just SQL text); default width 100→50; dropping the `* width` re-multiply; disabling the conflicting-width check; accepting width 0; breaking the underscore-spelling derivation; lenient unclosed-paren parsing; `Objects.equals`→`!=` on boxed widths; width-boundary drift `< 1`→`< 2`; cross-field conflict-scoping loss; and deleting the `AggregateRow.group` inclusion annotation (fails the wire test).
- **Review panel ran twice** (three-to-four read-only lenses each round, each refuting its own findings before reporting), plus a **four-lens simplification pass** (reuse / simplification / efficiency / altitude, net −56 lines: the `GroupKind` restructure, factory/fixture/helper collapses, doc dedup) and a **comment pass** (current-purpose comments only; provenance lives in the log). A reviewer-found REST wire-contract bug on null group keys was confirmed by reproduction and fixed test-first (details above). Efficiency lens recorded one watch item: high-cardinality groupings trip the pre-existing totals second query more often than the old groupable fields did.

https://claude.ai/code/session_01GJ8tr6QSDNLb21VkcbkcKy


aaylward:
Finding 1 confirmed and fixed in f7583af — the reproduction matched your table byte for byte: the all-null group served `{"count":1}` with no `group`, and `{opponent_title: null, outcome: "win"}` served only the outcome.

Both demanded tests shipped, in `AggregateResponseWireTest` (mirrors `IndexResponseWireTest`: real `EmbeddedServer`, raw `HttpClient`, assertions on the raw body):

1. `aggregateResponseKeepsExplicitNullGroupKeysOnTheWire` — asserts `"group":{"opponent_elo":null}` on the wire, with the non-null control (`"group":{"opponent_elo":2400}`, a JSON number) in the same response.
2. `aggregateResponseKeepsNullDimensionAlongsideNonNullKeysOnTheWire` — asserts `"group":{"opponent_title":null,"outcome":"win"}` next to `"group":{"opponent_title":"FM","outcome":"win"}`.

Both failed on exactly the dropped-key bodies before the fix; the fix came after. It's `@JsonInclude(value = ALWAYS, content = ALWAYS)` scoped to `AggregateRow.group` only — `IndexResponseWireTest`'s null-bean-field omission pins still pass, and a mutation check deleting the annotation fails the wire test.

---

PR #1312, opened 2026-08-05
one_d4: serve the first-load query from a warmed in-memory cache

aaylward:
## What

First page load of 1d4.net always fires the same request — `{"query":"num.moves >= 0","limit":25,"offset":0}` — and paid a full parse/compile/DB round trip every time. This keeps that one response in memory so first paint skips the database:

- **`FirstPageCache`** — single-entry Caffeine `LoadingCache` (`expireAfterWrite` 60s, monotonic system `Ticker`; tests inject a `MutableTicker`). The loader is the one definition of how a snapshot is computed; `get()` loads on a cold/expired miss under Caffeine's per-key lock, so concurrent misses share one query. Warming is a `put` of the loader's result, deliberately not `LoadingCache.refresh` — refresh discards a reload whose entry expired mid-load, which would silently leave a slow tick cold. Caffeine is promoted from a transitive of error-prone to a direct pin (`3.2.2`) in `bazel/java.MODULE.bazel`, `maven_install.json` repinned.
- **`FirstPageWarmer`** — `@Scheduled` refresh every 30s. The schedule is the point, not an optimization: on a low-traffic site a lazily-populated cache misses for exactly the first visitor it exists for. The 60s time bound replaces write invalidation, which can't work here anyway — index workers on other JVMs write to the shared database.
- **`QueryExecutor`** — the controller's pipeline extracted so warmed responses are built by the same code a live request runs.
- **`idx_game_features_played_at`** — new migration index on `game_features(played_at DESC, game_url ASC)`, mirroring SqlCompiler's browse `ORDER BY` exactly. Without it every page — the warmed default, the page-2 prefetch the cache excludes, and every loosely-filtered query — was a full scan plus top-N sort; with it, a LIMIT-sized index walk. The migration test pins column order and directions, not mere existence.
- The first-load wire body lives in a shared fixture (`first_page_request.json`) pinned by the Java constants test, POSTed verbatim by the wire test, and read by `GamesView.test.tsx` — neither side can drift from the cache key by editing only its own literal. `GameFeatureRow.fromStore` defensively copies the occurrences map now that rows outlive a request inside the shared snapshot.

## Review panel

Ran the 4-lens read-only panel (correctness / data-access & resources / tests-docs-CI / altitude). No blocking defects; every surviving finding was fixed with tests or filed as follow-up (#1313). Two rounds of external review (cursor bot) were addressed the same way: compiled-query pin in the fake store, the cross-language fixture gate, and the refresh-discards-slow-reload fix — each verified by mutation check.

Mutation checks: 18/18 killed across `FirstPageWarmer` (wrong request warmed — including same-pagination-different-query — slow schedule, no-op tick), `QueryExecutor` (limit/offset swap, count), `QueryController` (cache bypassed, bypass writes cache), `FirstPageCache` (each `matches()` guard, constants drifting from the fixture, warm-via-refresh regression), and `GamesView.tsx` (default query/page size).

## Verification

- `bazel test //domains/games/apis/one_d4/...` — 56/56 pass
- `FirstPageWarmupTest` (embedded server, timing-sensitive) — stable at `--runs_per_test=15`
- 1d4_web: `npm run typecheck && npm test && npm run build` — 120/120 pass
- `scripts/format-java` + buildifier run
- Not exercised locally: the Postgres-gated suites (`pg_db_tests`) — no `PG_TEST_DB_URL` in the sandbox. The new `CREATE INDEX` is dialect-neutral and CI's `postgres:18` job exercises the migration. Verified via the sandbox git-override workaround for proxy-blocked archives (plus a local `smithy_cpp` override); no lockfile churn committed beyond the intentional caffeine repin.

## Follow-ups (tracked in #1313)

Measured-first ordering: read the `POST /v1/query` route histograms, then response compression on `api.1d4.net`, dropping `pgn` from the list payload, frontend bundle splitting, a cache hit/miss metric, and a statement timeout on the query path. (The browse index was originally on this list and moved into this PR.)


aaylward:
Findings 1–2 closed with tests in 272b635:

1. `FakeGameFeatureStore` now records the compiled query (the `AggregateControllerTest` `lastCompiled` pattern), and the warmer test asserts it equals the compilation of the default request — the wrong-query-right-pagination mutation (`"white_elo >= 3000"`, limit 25, offset 0) is now killed; verified with `scripts/mutation-check`.
2. The wire body moved into a shared fixture, `one_d4/src/test/resources/first_page_request.json`: `FirstPageCacheTest` pins the Java constants to it, `FirstPageWarmupTest` POSTs it verbatim, and `GamesView.test.tsx` reads the same file. Mutating either side's constants (`DEFAULT_QUERY`/`DEFAULT_LIMIT` in Java, `DEFAULT_QUERY`/`DEFAULT_PAGE_SIZE` in the frontend) now fails that side's own CI job — all four mutations verified killed.

On 3: agreed the hand-rolled slot expressed the same policy, but replacing it with Caffeine was an explicit owner decision on this PR, so it stays.

---

aaylward:
Good catch — confirmed and fixed in f52a191. `refreshNow()` now `put`s the loader's result instead of calling `LoadingCache.refresh`, so a successful warm always installs even when the prior entry expired mid-load; failure still keeps the last good snapshot and logs. `get()` keeps the loader and its single-flight. The expiring-mid-load scenario is pinned by `refreshNow_slowTickWhoseEntryExpiresMidLoadStillInstallsTheResult` (the fake store advances the test ticker past `MAX_AGE` inside the load), and `scripts/mutation-check` verifies the test fails if the body is mutated back to `refresh(KEY).join()`.

Docs nit also fixed in the same commit — API.md and the README now describe the miss path as load-through with shared cold misses. The single-flight concurrency test is left as-is per your "optional": it's library behavior, and a thread-race test would buy flake risk more than proof.

---

Issue #1313, opened 2026-08-05
one_d4 first-load latency: the levers the first-page cache doesn't pull


aaylward:
Follow-ups from the review panel on #1312 (first-page cache), extended by the holistic altitude pass over #1312 + #1314. The cache removes server think-time for the default request, but the panel's altitude/resources lenses found bigger, unmeasured terms in the same first-load chain. Worth taking in roughly this order — and step 0 first, so the rest is aimed by data rather than plausibility.

- [ ] **0. Measure before optimizing further.** The `http_server_*` per-route histograms landed just before #1312; read the `POST /v1/query` latency split (and browser-side timing against `api.1d4.net`) to see which of the terms below actually dominates first paint. Include client-side **bundle parse/execute time** here — see the demotion note on item 4.
- [ ] **1. Response compression on `api.1d4.net`.** The Caddyfile has `encode` only in the `r3dr.net` block; the API block has none, and one_d4 has no Micronaut-side compression. Every first load ships ~50–150KB of JSON — including 25 full PGNs — uncompressed. Likely a one-liner: `encode zstd gzip`.
- [ ] **2. Drop `pgn` from the list response (or add a detail endpoint).** The games list never renders `pgn`; only `GameDetailPanel` does, on click. It's the bulk of the payload bytes — and the altitude pass adds a second, server-side reason: every 30s warmer refresh detoasts 25 full PGNs on a 0.25-CPU Postgres, so dropping `pgn` from the list projection cuts the warmer's steady-state DB load ~10× too, not just the wire bytes. Note the detail panel currently renders from the list payload, so this needs a fetch-on-click or a slimmer list projection plus detail lookup. Once landed, the warmer tick can also relax (e.g. 2min tick / 5min MAX_AGE) if a few minutes of first-page staleness is acceptable.
- [x] **3. Index behind the browse ordering.** Done in #1312 (commit 77c1abd): `idx_game_features_played_at ON game_features(played_at DESC, game_url ASC)` in Migration, mirroring SqlCompiler's exact `ORDER BY`, with a migration test asserting column order and directions rather than mere existence.
- [ ] **4. Split the frontend bundle.** *Demoted by the altitude pass:* the 394KB chunk ships as ~110–120KB brotli via Cloudflare, so wire size is not the lever it looked like — the real question is parse/execute time on low-end devices, which item 0 should measure before any `React.lazy` work. (`chess.js` and `react-chessboard` remain the obvious lazy candidates if the measurement says so.)
[...]

PR #1314, opened 2026-08-06
one_d4: bound read-query execution with a statement timeout

aaylward:
## What

Items 6–9 of #1313. Nothing bounded query *execution* — HikariCP's `connectionTimeout` bounds pool checkout only — so a statement wedged on a lock wait ran forever, and every unattended loop that can't re-enter while a run is in flight stayed wedged until a restart: the first-page warmer's `@Scheduled` slot, the dispatch poller's single claiming thread, and the hourly retention sweep.

- **`StatementTimeouts`** — the one entry point (`withStatementTimeout`, plus `inTransactionWithTimeout` sharing the same bounded core) that applies a bound via the handle's statement config and clears the session on the way out. Bounds are structural, not a call each statement must remember; a mutation routing a read around the entry point is killed.
- **Serving reads: 10s** (`query`, `aggregate`, `aggregateTotals`, `queryOccurrences` via `withReadHandle`, and the `GET /v1/index` status reads `listRecent` + `findPeriodsForPlayers`). The binding constraint is `FirstPageCache.MAX_AGE` (60s) — a warmer tick runs **two** bounded reads, and `fixedDelay + 2×timeout < MAX_AGE` is pinned by a test.
- **Dispatch poller: 10s** on `claimNext`'s candidate scan — it runs every few seconds on a single dedicated thread whose loop never returns while a statement is in flight; unbounded, one wedged scan stopped the instance claiming any work. The claim UPDATEs that follow stay unbounded (pinned).
- **Retention sweep: 120s** on all three `deleteOlderThan`s and the stale-request reclaim transaction — `RetentionWorker` shares the warmer's scheduled pool with no lease or interrupt machinery, and a truncated idempotent sweep re-runs in an hour, so the "writes stay unbounded" rationale (which holds for the index worker's flushes and `fetchForReanalysis`) never reached it.
- **`socketTimeout=150` default for Postgres URLs** in `DataSourceFactory` — in code, versioned and tested, rather than an instruction against unversioned per-host config. Statement cancellation travels over the network; on a TCP black hole only a socket timeout gets the connection back. The value must exceed the sweep bound (a long DELETE is silent on the socket until it finishes) — pinned — and an explicit URL value always wins; H2 URLs are excluded (H2 rejects unknown connection properties).
- **H2 session-leak containment**: H2 scopes `setQueryTimeout` to the connection's session (verified in driver bytecode) and Hikari doesn't reset it on checkin, so without cleanup a bound would leak to every later statement on the pooled connection — including the deliberately unbounded writes. The entry point clears the session in a `finally`, through the JDBC API (raw `SET QUERY_TIMEOUT 0` leaves H2's client cache stale), logging rather than throwing so cleanup failure can't mask the timeout it exists to surface. On Postgres the clear is a pure client-side no-op, so it runs unconditionally and the helper carries no dialect knowledge. The leak was surfaced by mutation checking — probes were passing by inheriting an earlier read's leaked session value.

## Holistic altitude pass (final commit)

A three-lens review over the whole two-PR arc (#1312 + this) closed the gaps it found:

- **Two behaviors were argued in prose but unpinned — both now tested and mutation-verified**: the dispatch poller surviving a `claimNext` failure and claiming on a later poll (`IndexWorkerLifecycleTest`), and a failing/cancelled read surfacing as an error rather than a silently empty page, on both the live path and the cache's cold-miss loader (`QueryControllerTest`). The retention sweep's survive-store-failure contract got the same treatment (`RetentionWorkerTest`).
- **`GET /v1/index` reads were the last unbounded serving path** — `listRecent` and `findPeriodsForPlayers` now carry the 10s bound, pinned in `StatementTimeoutsTest` (removing either bound is a killed mutation).
- **Consolidation**: `StatementTimeouts` is the single home of the policy (vestigial `GameFeatureDao` timeout field/constructor removed); all mechanism tests live in `StatementTimeoutsTest` with the literals 10/120 in exactly one place; `PostgresReadTimeoutTest` reduced to pure `pg_sleep` mechanism (no migrations or tables); a stale javadoc, a subsumed frontend test, and a vacuous sweep-autocommit assertion removed; `GamesView.test.tsx` now reads the cross-language first-page fixture directly.

Posture findings that belong to future work (username-search indexes, `indexed_at` index for the sweep, serve-stale-on-error cache, bounded-loop outcome metrics, the 150s socket cliff on reanalysis) are filed on #1313 as items 10–13 and re-scopes of 2/4/5.

## Tests

- **H2 behavioral cancellation** (`StatementTimeoutsTest`): per-row sleeping UDF over 400 rows (H2 checks the timeout cooperatively every N rows) killed near a 1s injected timeout, with an error-path probe asserting the timed-out read left no timeout on the pooled connection, and a fast positive twin.
- **Real-Postgres suite** (`PostgresReadTimeoutTest`, new, in `pg_db_tests`): `pg_sleep` cancelled at the timeout, a behavioral no-leak probe — after a 1s-bounded read, an unbounded 2s `pg_sleep` must complete, and `SHOW statement_timeout` pins no server-side session setting was left — and the socket-timeout default pinned via `Connection.getNetworkTimeout` on a real connection. Env-gated on `PG_TEST_DB_URL` (CI supplies `postgres:18`); **also run locally against a real PostgreSQL 16** — see Verification.
- **`StatementTimeoutsTest`** (new): probes the poller scan (10, and the claim UPDATEs at 0), the status-path reads, the reclaim transaction, and all three sweep deletes (120) on the real JDBC statements via `SqlLogger`, plus session-leak probes on the success and error paths. `DataSourceFactoryTest` (new) pins the socket-timeout policy: default applied, explicit URL wins, H2 excluded, value exceeds the sweep bound.
- Probe assertions use literals so mutating a constant can't mutate the expectation; timing-sensitive tests re-run at `--runs_per_test=15`.

## Review panel

Ran the 4-lens read-only panel on the initial commit. Correctness: one finding (finally-throw replacing the primary exception) — fixed. Resources: no defects; confirmed from bytecode that Hikari never plants its own session timeout on H2 and pgjdbc uses one shared JVM-wide cancel timer. Tests/docs: found an escaped mutation (reset moved from `finally` to `try` survived) — now killed by the error-path probe, verified by applying the mutation and watching it fail; found the original PG leak probe was vacuous — replaced with the behavioral probe. Altitude: right mechanism (it attacked `connectionInitSql`, URL `options`, and HTTP-layer bounds; each failed on merits), but the boundary was class-shaped rather than risk-shaped — its three boundary findings are the poller/sweep/socket work now included in this PR. The closing three-lens holistic pass is summarized above.

Mutation checks: all killed — per-path timeout removal, constant drift (both values), session-reset removal, reads bypassing the entry point, the escaped finally-placement mutation, poller-bound removal, socket-timeout policy inversions, (against real Postgres) removing the bound from `withStatementTimeout`, and from the altitude pass: poll-loop catch removal, retention catch removal, read-failure swallowed into an empty page, and status-path bound removal on both DAOs.

## Verification

- `bazel test //domains/games/apis/one_d4/...` — 60/60 pass, **including `pg_db_tests` against a real local PostgreSQL 16.13** (`PG_TEST_DB_URL` exported): zero skips, `PostgresReadTimeoutTest` stable over `--runs_per_test=15` (real cancellations). CI additionally runs the same suites against `postgres:18`.
- `1d4_web`: 119 vitest tests pass, `tsc --noEmit` clean.
- `scripts/format-java` + buildifier run
- Built via the sandbox git-override workaround for proxy-blocked archives; no lockfile churn committed.

Docs updated in the same PR: README (socketTimeout default and override rule), API.md (deterministic 500 on a query exceeding the read timeout), ROADMAP (`SET statement_timeout` item marked done in this form and why not that one), and the warmer's now-false "a hung query blocks future ticks" comment corrected.

### DIFF
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -580,6 +580,7 @@ java_test_suite(
     srcs = [
         "src/test/java/com/muchq/games/one_d4/db/PostgresAggregateCompatTest.java",
         "src/test/java/com/muchq/games/one_d4/db/PostgresConcurrentWriteTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/PostgresReadTimeoutTest.java",
     ],
     env_inherit = ["PG_TEST_DB_URL"],
     # One shared scratch database: concurrent DB tests drop each other's schemas.
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -698,7 +698,9 @@ java_test_suite(
 
 java_test_suite(
     name = "http_integration_tests",
-    size = "small",
+    # medium (300s), not small: FirstPageWarmupTest boots an embedded server per method plus a
+    # polling warm-up wait, which sits too close to small's 60s ceiling on a loaded CI machine.
+    size = "medium",
     srcs = [
         "src/test/java/com/muchq/games/one_d4/api/AggregateResponseWireTest.java",
         "src/test/java/com/muchq/games/one_d4/api/ContentTypeHandlingTest.java",
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -78,6 +78,7 @@ java_library(
         "src/main/java/com/muchq/games/one_d4/db/IndexingRequestStore.java",
         "src/main/java/com/muchq/games/one_d4/db/Migration.java",
         "src/main/java/com/muchq/games/one_d4/db/RetentionPolicy.java",
+        "src/main/java/com/muchq/games/one_d4/db/StatementTimeouts.java",
     ],
     visibility = ["//visibility:public"],
     deps = [
@@ -502,10 +503,12 @@ java_test_suite(
     size = "small",
     srcs = [
         "src/test/java/com/muchq/games/one_d4/db/ConcurrentFlushTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/DataSourceFactoryTest.java",
         "src/test/java/com/muchq/games/one_d4/db/GameFeatureDaoTest.java",
         "src/test/java/com/muchq/games/one_d4/db/IndexedPeriodDaoTest.java",
         "src/test/java/com/muchq/games/one_d4/db/IndexingRequestDaoTest.java",
         "src/test/java/com/muchq/games/one_d4/db/MigrationTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/StatementTimeoutsTest.java",
         "src/test/java/com/muchq/games/one_d4/db/TableDispatchTest.java",
     ],
     deps = [
--- a/domains/games/apis/one_d4/README.md
+++ b/domains/games/apis/one_d4/README.md
@@ -141,8 +141,13 @@ line) instead of using an environment variable. The app resolves config in this 
 
 For Neon, the URL looks like:
 ```
-jdbc:postgresql://ep-xxx.us-east-2.aws.neon.tech/dbname?user=xxx&password=xxx&sslmode=require
+jdbc:postgresql://ep-xxx.us-east-2.aws.neon.tech/dbname?user=xxx&password=xxx&sslmode=require&socketTimeout=30
 ```
+
+Include `socketTimeout` (seconds) on any remote Postgres URL. The application bounds read-query
+*execution* with a 10s JDBC statement timeout, but that cancellation travels over the network — on
+a dead network only a driver-level socket timeout gets the connection back. Without it, a TCP
+black hole can still hang a query indefinitely.
 
 The server starts on **port 8080**. Then index a player and query:
 
--- a/domains/games/apis/one_d4/README.md
+++ b/domains/games/apis/one_d4/README.md
@@ -141,13 +141,15 @@ line) instead of using an environment variable. The app resolves config in this 
 
 For Neon, the URL looks like:
 ```
-jdbc:postgresql://ep-xxx.us-east-2.aws.neon.tech/dbname?user=xxx&password=xxx&sslmode=require&socketTimeout=30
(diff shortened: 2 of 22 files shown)


aaylward:
All six findings addressed in 2ec030e:

1. **`reclaimStale` bounded** at the sweep timeout via a new `StatementTimeouts.inTransactionWithTimeout` (the three settle UPDATEs stay one transaction), pinned by a probe asserting every statement in the transaction carries the bound — and verified against real local Postgres.
2. **ROADMAP** no longer says socketTimeout "remains deployment config"; it now describes the in-code default with URL override.
3. **`fetchForReanalysis` exclusion pinned** — a probe asserts its statements carry no timeout; a mute wrap in `withReadHandle` is a killed mutation.
4. **API.md** now says the second aggregate statement runs only when truncation triggers the totals query.
5. **`create()` wiring probed** — `hikariConfig(url)` is a package-private seam and the test asserts the built config's driver properties (pg: `socketTimeout=150`, H2: absent); `defaultSocketTimeout` going unapplied is a killed mutation.
6. **`claimNext` probe de-vacuoused** — asserts the claim succeeded and statements followed the scan before checking they're unbounded.

Both nits taken as documentation: the per-statement (not per-tick) semantics of `RETENTION_SWEEP_SECONDS` are now in its javadoc with the ~4-window worst case; the `contains("socketTimeout=")` check stays as-is — it can only false-positive into *not* overriding an operator's URL, which is the safe direction.

All new pins verified by mutation check; full suite including the PG-gated tests run green against a local PostgreSQL 16.

---

aaylward:
Both remaining nits closed in bb99bfd:

1. **Claim-UPDATE exclusion pin restored** — `claimNextsCandidateScanIsBounded` again asserts the statements after the scan are non-empty and all unbounded; the wrap-all-of-claimNext-in-the-serving-bound mutation is killed.
2. **`socketTimeout` check anchored at a parameter boundary** (`?`/`&` prefix) so the substring inside another parameter's value can't suppress the default. One honest note: the first version of the pin used a camelCase `fooSocketTimeout=` URL and *survived* the regression mutation — case-sensitive `contains` never matched it, so the realistic false positive is the value-substring case the shipped test now uses. Mutation killed.

---

PR #1317, opened 2026-08-07
one_d4: expression indexes for username search + the retention-delete index (#1313 items 10–11)

aaylward:
## What

Items 10 **and 11** of #1313 — the two index follow-ups from the holistic altitude pass.

### Item 10: username search (expression indexes)

Two compiler paths emit the same case-folded username predicate, `LOWER(white_username) = LOWER(?)` OR'd across sides: the browse UI's username search (`white.username = "x" OR black.username = "x"` through the STRING_COLUMNS equality branch — the highest-traffic consumer) and the participation guard every perspective-field query runs. No plain column index can serve either — both sides are case-folded — so this was the full-table scan on the player-search path, and with a 5-connection pool, five concurrent player searches held every connection for the full 10s serving-read bound from #1314.

- **Postgres**: two expression indexes, `idx_game_features_white_username ON game_features(LOWER(white_username))` and the black twin. One per side, not a composite: an OR across two columns is answered by a BitmapOr of two independent scans, and a BitmapOr's output is unordered, so a composite with `played_at` couldn't skip the sort either — it would only double index weight on the hottest-write table.
- **H2**: plain-column indexes under the same names — pure write cost there (a plain index can't serve a `LOWER(...)` predicate on H2 either), carried for migration-path parity; the one H2 deployment (the MCP server's boot-scoped in-memory store) makes both costs negligible.

### Item 11: the retention-delete index

`DELETE FROM game_features WHERE indexed_at < ?` runs hourly under the 120s sweep bound from #1314; unindexed, a sweep that hit the bound rolled back having made **no forward progress** and retried the identical table scan an hour later — a ratchet that never advances. The delete is one statement, so a cancel always rolls the whole pass back; what `idx_game_features_indexed_at` (plain b-tree, dialect-neutral) changes is the **cost** side — the steady-state delete walks straight to the expired rows and *finishes* inside the bound. (`LIMIT`-batched deletes remain the next tool if a backlog ever outgrows the bound; deferred.)

## Tests

- **`PostgresPlayerIndexTest`** (new, in `pg_db_tests`): `EXPLAIN` under `enable_seqscan = off` must name both username indexes and contain no Seq Scan — for the guard's compiled query *and* the UI's browse-search query. Reachability, deliberately not planner choice (the javadoc explains why the DAO's `LIMIT`/`OFFSET` suffix is excluded). A behavioral twin pins case-insensitive matching on both colors through the real DAO, on both compiler paths.
- **`PostgresRetentionIndexTest`** (new, in `pg_db_tests`): the sweep's exact DELETE shape must reach `idx_game_features_indexed_at`, same technique.
- **`MigrationTest`**: index presence *and columns* on H2 (a stand-in redirected at the wrong column kept the name the first assertion looked for), plus the compiled predicate shape for both username paths, plus the retention index's column.
- **`SqlCompilerTest`**: a `player` with no perspective field in play adds **no** participation guard, pinned in both directions — documented contract (the guard repairs perspective resolution, it doesn't scope), pinned because the aggregate half is a footgun (item 14 on #1313).

Mutation checks, all killed against real Postgres 16 / H2: white index losing its `LOWER`; black index redirected at the white column; the STRING_COLUMNS compiler branch losing its `LOWER`; the H2 stand-in redirected at the wrong column; the retention index redirected at `played_at`; and the retention index's creation dropped from `run()`.

## Review panel

3-lens read-only panel (correctness, tests, altitude) on item 10, each lens self-refuting; all survivors applied — including the altitude lens's coverage-hole find (the browse path was unpinned) and cursor[bot]'s re-review nits across both rounds (javadoc narrowing, `doesNotContain("Seq Scan")`, H2 column pins, the first-deploy SHARE-lock ops note, and the truncated-pass semantics correction above). Item 11 follows the exact pattern the panel settled for item 10.

Housekeeping from the tests lens: the libpq→pgjdbc URL helper, previously copied into four PG suites (about to be five), now lives once in `PgTestUrls` under `:test_db`; all five suites use it.

## Verification

- Full `//domains/games/apis/one_d4/...` sweep (62/62) + `chessql` compiler suite green, including all five PG-gated suites against a real local PostgreSQL 16 (zero skips). CI additionally runs `postgres:18`.
- `scripts/format-java` + buildifier run; no lockfile churn.

## Follow-ups on #1313

Item 14 (the unguarded-aggregate footgun: `player` + plain filter on `/v1/aggregate` aggregates the whole corpus presented as the player's) and item 15 (stale README H2-storage claim) remain filed there.

### DIFF
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -583,6 +583,7 @@ java_test_suite(
     srcs = [
         "src/test/java/com/muchq/games/one_d4/db/PostgresAggregateCompatTest.java",
         "src/test/java/com/muchq/games/one_d4/db/PostgresConcurrentWriteTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/PostgresPlayerIndexTest.java",
         "src/test/java/com/muchq/games/one_d4/db/PostgresReadTimeoutTest.java",
     ],
     env_inherit = ["PG_TEST_DB_URL"],
--- a/domains/games/apis/one_d4/BUILD.bazel
+++ b/domains/games/apis/one_d4/BUILD.bazel
@@ -585,6 +585,7 @@ java_test_suite(
         "src/test/java/com/muchq/games/one_d4/db/PostgresConcurrentWriteTest.java",
         "src/test/java/com/muchq/games/one_d4/db/PostgresPlayerIndexTest.java",
         "src/test/java/com/muchq/games/one_d4/db/PostgresReadTimeoutTest.java",
+        "src/test/java/com/muchq/games/one_d4/db/PostgresRetentionIndexTest.java",
     ],
     env_inherit = ["PG_TEST_DB_URL"],
     # One shared scratch database: concurrent DB tests drop each other's schemas.
@@ -596,6 +597,7 @@ java_test_suite(
         ":db",
         ":dto",
         ":model",
+        ":test_db",
         "//domains/games/libs/chessql:compiler",
         "//domains/games/libs/chessql:parser",
         artifact("org.assertj:assertj-core"),
@@ -608,6 +610,7 @@ java_library(
     name = "test_db",
     testonly = True,
     srcs = [
+        "src/test/java/com/muchq/games/one_d4/db/PgTestUrls.java",
         "src/test/java/com/muchq/games/one_d4/db/TestDb.java",
     ],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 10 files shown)


aaylward:
All four nits taken in df05cfc:

1. **Class javadoc narrowed** — `MigrationTest` owns compiler-shape drift on H2 now; this suite's unique claim is the index side of the contract.
2. **Plan pins assert `doesNotContain("Seq Scan")`** on both paths, so the assertions say what the javadoc sells.
3. **H2 stand-ins pinned to their columns**, not just names — and the redirect-at-the-wrong-column mutation is killed (it survived the name-only assertion, as predicted).
4. **First-deploy ops note** added to the Migration javadoc, including why `CONCURRENTLY + IF NOT EXISTS` would be the worse trade.

All suites re-run green against local Postgres 16 (the intervening red run was the sandbox's postgres dying, not the diff).

---

Issue #1194, opened 2026-07-23
golf_hub persistence: restart survival now, horizontal scaling later


aaylward:
Research + plan for taking golf_hub (#1187) from all-in-memory to a persistence story that enables horizontal scaling. Written against the phase-4 deploy shape (single container behind Caddy, compose stack, one_d4's postgres 18 already running).

## What actually blocks scaling out

golf_hub holds four kinds of state, and they fail differently when a second instance appears:

1. **Credentials** (`TicketVault`): single-use tickets (30s) and resume tokens (24h), in-memory. With N instances, `POST /games/v2/session` can mint on instance A and the WebSocket dial can land on B, which has never seen the ticket. Breaks immediately, even with sticky LB, because the mint is a separate HTTP request from the upgrade.
2. **Rooms, membership, stats** (`HubHandler::rooms_` etc.): small relational data. A room created on A is invisible on B — permalinked game codes stop working the moment there are two instances.
3. **Game state** (`GameEntry::state`, the engine's `GameState`): the full truth including deck order and unrevealed cards. Immutable value type with pure transitions — no serialization exists today.
4. **Live fan-out** (`SessionRegistry`): sockets are process-local. A move processed where the game lives must reach players whose sockets live elsewhere. This is the only piece that needs a *bus*, not just a store.

Two adjacent facts worth keeping in frame:

- **Restarts already lose everything.** Every deploy kills live games and forgets all resume tokens (v1 has the same flaw — its JWT secret rotates on restart). Persistence buys restart survival long before a second instance exists, and that's the more immediate win: deploys stop being player-visible events.
- **The reconnect story gets better, not worse.** Today grace timers die with the process. With state external, a player whose instance died can resume anywhere.

### A. Postgres as system of record + `LISTEN/NOTIFY` fan-out — recommended

Schema sketch (own database + role on the existing postgres instance; migrations via a small in-service runner, the pattern one_d4 already tests):

- `resume_tokens(token_hash, player_id, expires_at)`; `tickets(ticket_hash, player_id, expires_at)` — spend = single-row `DELETE ... RETURNING`, which is exactly the single-use semantic, now cross-instance.
- `rooms(room_id)`, `room_members(room_id, player_id, connected, games_played, games_won, total_score)`.
[...]

PR #1320, opened 2026-08-07
one_d4 + 1d4.net: refuse unscoped player aggregates, correct the storage docs, add a /mcp page (#1313 items 14–15, #1321)

aaylward:
Three issues, all one_d4-adjacent: two backend/doc corrections from #1313 and the new `/mcp` page from #1321.

## #1313 item 14: the unguarded-aggregate footgun

A `player` only narrows a query through the participation guard, and the guard only fires when a perspective field is in play — it exists to repair the `CASE` expressions ("not white" means "black"), not to filter. On `/v1/query` that is harmless: the rows carry usernames, so a caller who expected one player's games sees strangers in the first result. An aggregate has no such tell:

```
player: hikaru, query: 'num.moves >= 0', group_by: ["opening_family"]
```

— the natural MCP call for "hikaru's most common openings" — returned **the whole corpus's** openings under a heading that reads as hikaru's, and nothing in the response revealed it. Now rejected (400).

**Where the rule lives.** In `SqlCompiler`, at both aggregate entry points — not the API validator. `IndexerFacade` (the MCP path, where an assistant phrases the request most naturally) never touches `AggregateController`, and the totals query runs separately on the truncation path. The compiler is the only layer both callers share, and it already knows `perspective.used`, which an API-layer check would re-derive by duplicating the perspective roster.

**What still compiles.** A filter that *cannot match another player's game* counts as scoping. That test is checked, not assumed, which makes the cases deliberately asymmetric: `AND` scopes if any conjunct does, `OR` only if every branch does, `NOT` never does, only `=` pins a value, an `IN` list scopes only when every alternative is the player, and the name must *be* the named player (case-folded). Naming no player is untouched. `/v1/query` keeps its permissive semantics, now stated in CHESSQL.md rather than only pinned in a test.

**Blast radius: empty.** Every existing `compileAggregate(..., player)` call site uses a perspective field; `1d4_web` never calls `/v1/aggregate`.

## #1313 item 15: storage docs

The one_d4 README described the deployed indexer as H2 file storage at `/data/indexer` with copy-the-`.mv.db` instructions. It has run PostgreSQL since #1194 (`one_d4_postgres`, volume `one_d4_pgdata`, URL from `/etc/one_d4/db_config`) — replaced with the `psql`/`pg_dump` path against the user and database Compose actually creates. ROADMAP's disk-cap item justified itself with the `one_d4_data` volume that no longer exists and a "once the indexer uses PostgreSQL" note that has since happened; rewritten as what's left (a host-level quota on `one_d4_pgdata`).

The mcpserver README now records that its in-process indexer really does take the in-memory H2 default — `INDEXER_DB_URL` is the only source `McpModule` reads, and Compose sets none — so its index is per-process and empty at boot. Pointing it at the shared corpus would give MCP indexing tools write access to production data: a deliberate deployment decision, deferred rather than quietly made.

## #1321: the /mcp page

The issue asked that three questions be **settled before writing the page**. Answered by connecting to the deployed endpoint rather than reading the code:

- **Transport.** Plain JSON-RPC 2.0 over a single `POST /mcp`. No SSE stream (`GET /mcp` answers 200 with an empty body), no `Mcp-Session-Id`, and `notifications/initialized` is unimplemented — it answers `-32601`, to a notification that per JSON-RPC should draw no response at all. **No standard MCP client transport connects as deployed.** The page says so plainly instead of shipping a config that would not work; the `curl` recipe it gives is verified against the live server.
- **Auth.** An unauthenticated `tools/list` returns 200 with results, so the deployment runs open. README and Compose agree — no mismatch to discover on a public page.
- **CORS.** No `access-control-*` headers on `mcp.1d4.net`, confirming the page must be documentation, not an interactive client.

The page covers the roster, the index-first ordering (the one thing a newcomer gets wrong), the connection caveat, a worked example, and a pointer to `/query` for ChessQL. Route and nav link follow the existing three exactly; `wrangler.toml` already handles SPA deep links and `1d4_web` has no Bazel targets, so neither needed work.

**The table cannot rot silently** — the issue's own "interesting test". `mcp_tools.json` is the contract: `McpToolRegistryContractTest` asserts the server's registry matches it, built from `McpModule`'s own factory rather than a copy of the list, and `McpView.test.tsx` asserts the page matches the same file. Verified by adding a phantom tool and watching both ends fail.

## Tests

- **Aggregate refusal** at three levels, each pinning that no query ran: compiler (`SqlCompilerTest` — a table of five refused shapes, each with the reason it admits strangers), endpoint (`AggregateControllerTest`), MCP facade (`IndexerFacadeAggregateTest` — the path the endpoint test cannot reach). **Six mutations, six kills**: dropping the check, `NOT` recursing, the `=` check, the value check, `OR`→`anyMatch`, `AND`→`allMatch`, `IN`→`anyMatch`.
- **/mcp page**: roster matches the server contract, every row carries a summary, the index-first ordering and the transport caveat are present, and the route renders with its nav link active.

## Review panel

Two read-only lenses on the aggregate work, each self-refuting; both found real defects, both applied.

- **Correctness** found the original predicate weaker than the guarantee — and *pinned that way*: `NOT white.username = "magnus"` compiled and returned every blitz game magnus didn't play as White, under hikaru's heading, with a test asserting it was fine.
- Mutation testing then caught a second-order problem the (passing) new tests hid: every negative case used *magnus*, so the value check alone rejected them and the `NOT`/`!=` logic was unpinned. The killing cases name the player himself.
- **Altitude/tests** confirmed the compiler is the right layer, confirmed throwing over warning-or-auto-scoping, and found the tightening had shipped without its doc sweep — three places still described the loose rule, API.md's 400 table was missing the condition, ROADMAP had the stale volume, and one test label asserted the opposite of what it observed. All fixed.

## Verification

76/76 Java tests across `chessql`, `one_d4`, `mcpserver`; 124/124 frontend tests, `tsc` clean, Vite build clean. `scripts/format-java` + buildifier run; no lockfile churn.

## Deferred, noted rather than half-done

A standalone ChessQL route for `/query` and `/mcp` to share (#1321 suggests it; independently scoped), and the transport gap itself — worth its own issue now that it's established rather than suspected.

### DIFF
--- a/domains/games/apis/mcpserver/BUILD.bazel
+++ b/domains/games/apis/mcpserver/BUILD.bazel
@@ -146,6 +146,13 @@ java_binary(
 
 linux_oci_java(bin_name = "mcpserver")
 
+java_library(
+    name = "mcp_tools_fixture",
+    testonly = True,
+    resource_strip_prefix = "domains/games/apis/mcpserver/src/test/resources",
+    resources = ["src/test/resources/mcp_tools.json"],
+)
+
 java_test_suite(
     name = "tools_tests",
     size = "small",
@@ -156,9 +163,12 @@ java_test_suite(
         "src/test/java/com/muchq/games/mcpserver/tools/ChessComStatsToolTest.java",
         "src/test/java/com/muchq/games/mcpserver/tools/IndexerFacadeAggregateTest.java",
         "src/test/java/com/muchq/games/mcpserver/tools/IndexerToolsTest.java",
+        "src/test/java/com/muchq/games/mcpserver/tools/McpToolRegistryContractTest.java",
         "src/test/java/com/muchq/games/mcpserver/tools/ServerTimeToolTest.java",
     ],
     deps = [
+        ":mcp_tools_fixture",
+        ":module",
         ":tools",
         "//domains/games/apis/one_d4:db",
         "//domains/games/apis/one_d4:dto",
--- a/domains/games/apis/mcpserver/README.md
+++ b/domains/games/apis/mcpserver/README.md
@@ -254,6 +254,14 @@ The indexer engine, database, and worker are embedded in this process (Option A 
 `one_d4/.../docs/MCP_INTEGRATION.md`). By default the index lives in in-memory H2; set
 `INDEXER_DB_URL` to point at a durable database.
 
+**The deployed MCP server takes that default.** `INDEXER_DB_URL` is the only source `McpModule`
+reads — there is no `/etc/…/db_config` fallback like `one_d4`'s — and the Compose service sets no
+such variable, so its index is in-memory H2, scoped to the process: empty at boot, discarded on
+restart, and entirely separate from the `one_d4` service's PostgreSQL corpus. Indexing tools work
+here, but each container starts from nothing and re-indexes on demand. Pointing this at the shared
+database is a deliberate deployment change (it would write to the production corpus), not a
+missing setting to paste in.
+
 - `index_chess_games` — index a player's games. Required: `username`, `platform`
   (`chess.com`), `start_month`/`end_month` (YYYY-MM). Optional: `exclude_bullet`, and
   `skip_cache` to refetch already-indexed months (refreshing stored rows, e.g. backfilling
--- a/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/tools/AggregateGamesTool.java
+++ b/domains/games/apis/mcpserver/src/main/java/com/muchq/games/mcpserver/tools/AggregateGamesTool.java
@@ -32,8 +32,10 @@ public class AggregateGamesTool implements McpTool {
   public String getDescription() {
     return "Count indexed games grouped by one or more fields, filtered by a ChessQL query"
         + " (index first with index_chess_games). This answers questions like 'most popular"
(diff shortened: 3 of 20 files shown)


aaylward:
Both blockers fixed in 1f3a76c — I verified each against the code before changing the copy, and both were right.

**1. The index is separate, and the page didn't say so.** Now its own panel (“Its index is its own”), placed above the connection details: in-memory, empty at boot, discarded on restart, not the corpus behind Games/`api.1d4.net`, with a pointer to the Index page for filling the corpus. Fair hit — this same PR documented the fact in the mcpserver README and then omitted it from the page most able to mislead about it.

**2. Sync vs async.** Confirmed: `IndexerFacade:61` calls `submitHybrid`, and `IndexRequestService.start` runs inline when `monthSpan <= 1` — so single-month is synchronous for the MCP path, while `IndexController:69` uses the async `submit` (which is where my copy's "returns a request id immediately" came from — right for REST, wrong for the audience this page is written for). The page now says a single month is indexed while you wait and a range is queued, the tool summaries match, and the worked example — one month — says "nothing to poll" instead of sending readers to poll a finished request.

Both pinned in the style of the transport caveat, so the next copy edit has to be deliberate. 126 frontend tests, typecheck and build clean.

---

PR #1322, opened 2026-08-07
metrics: give counters a zero baseline, and count bursty work over a day

aaylward:
## The report

> `muchq.com/metrics/one_d4` doesn't show 1d4 indexing metrics, but I indexed a bunch of games last night.

True, and worse than stale: the Indexing panel read **zero**, indistinguishable from a service that had never indexed anything.

## What I found

The emit path is fine. I confirmed that by triggering a real index run and watching `games_indexed` go 0 → 15.8 within 20 seconds — instrumentation, export, collector, Prometheus and dashboard all work. Two separate causes conspired instead, both general rather than one_d4's. Tracked as **#1323**.

**1. A counter's first event is invisible, permanently.**

A `CustomMetrics` counter only exists once something increments it, so the series is *born already carrying* the value of the first event it counted. `increase()` and `rate()` measure the change *between* samples — with no earlier sample to measure from, that first event shows no increase at all, and every sample after it is flat until something else happens. The counter isn't wrong (`games_indexed_total` really did read 1092); every dashboard built on it is.

That's not a corner case: it's what every process does after every deploy, so on every rail **the first run of anything has been uncounted**.

Dated from the deployed data rather than inferred — the reported run finished at **05:09:27**, and the `games_indexed` series first appears at **05:14**. Across seven days the only nonzero point on that metric is the run I triggered while investigating, visible precisely because the series already existed by then.

**2. The tiles count over five minutes.**

Right for serving traffic, wrong for work a person triggers: between asks the true rate is zero, so a 5m window answers "how busy is it right now?" when the question is "did the thing I ran work?".

## The fix

- **`CustomMetrics.defineCounter`** — declares a series so it exports as `0` from process start; the counter equivalent of the existing `defineDistribution`. Idempotent, and never resets a counter that has already counted.
- **`IndexWorker` declares its bounded series at construction**, not at the top of `process()` where the distribution bounds used to sit. A baseline only helps if it's exported *before* the increment it's a baseline for, and a run that starts and finishes between two export ticks would declare and increment inside the same interval — leaving the first run as invisible as it was. The worker bean is eager, so the baseline is minutes old before any request arrives.
- **Unbounded labels are left alone.** `motif_occurrences` is labelled per motif; declaring one series per possible value is how a metric becomes a cardinality problem, and the tile only reads their sum.
- **A `burstWindow` of 24h** for the indexing and motif volume tiles, joining the failure counters that already counted over a day. Named apart from `alarmWindow` deliberately: these are volumes, not alarms, and retuning how long a failure stays on screen shouldn't silently change how far back the counts reach.

**Scope note — the tiles were blind, not the charts.** The Trends `tsCounter` panels are unchanged (rate still over 5m, count still over the chart's step), and a 24h Trends range already showed an overnight run as a step-sized `increase` bump. What this fixes is the scalar tiles, which is what the report was about.

## Tests

- yodel: a declared counter exports at zero and makes the registry non-empty; counts normally once events arrive; survives re-declaration without losing its total; declaring one label set doesn't declare its siblings.
- one_d4: construction declares every bounded series at zero (asserting presence too, since summing an empty stream also yields 0 — the assertion would otherwise pass against a worker that declared nothing); and **a run emits no series construction didn't declare**, so an outcome added at an emit site without a declaration fails rather than quietly losing its first occurrence.
- prom_proxy: every Indexing/Motifs *volume* counter counts over `burstWindow`, checked exhaustively rather than by sample, with the constant itself pinned against a 12-hour floor. The run-outcome alarms are deliberately left to `TestRegistry_AlarmCountersCountOverALongWindow`, so the two constants stay independent — verified by mutation in both directions: shrinking `alarmWindow` survives the volumes test but is still killed by the suite.

**Mutations killed**: dropping `declareMetrics()`, dropping just the run-outcome declarations, making `defineCounter` a no-op, returning a tile to the 5m default, and shrinking each window constant.

## Verification

72/72 across `one_d4`, plus yodel and prom_proxy suites; `scripts/format-java` and gofmt clean.

## Note

Diagnosis required one real index request against production (`drawlya`, 2025-11, 15 games) — the service's own public operation, cleared by the 7-day retention sweep. It's the run visible in the 12:59 data point above.

### DIFF
--- a/domains/games/apis/one_d4/src/main/java/com/muchq/games/one_d4/worker/IndexWorker.java
+++ b/domains/games/apis/one_d4/src/main/java/com/muchq/games/one_d4/worker/IndexWorker.java
@@ -88,6 +88,19 @@ public class IndexWorker {
   static final String RUN_DURATION = "index_run_duration_micros";
 
   /**
+   * The label values the counters above can carry, so each series can be declared at startup and
+   * export a zero baseline. Bounded sets only: {@link #MOTIF_OCCURRENCES} is labelled by motif and
+   * left to appear on first use, since its baseline would be one series per motif for a tile that
+   * only ever reads their sum.
+   */
+  static final List<String> RUN_OUTCOMES =
+      List.of("completed", "failed", "interrupted", "lease_lost");
+
+  static final List<String> MONTH_RESULTS = List.of("indexed", "degraded", "cached", "empty");
+
+  static final List<String> ARCHIVE_FETCH_RESULTS = List.of("ok", "no_archive", "error");
+
+  /**
    * Bounds for {@link #RUN_DURATION}, in microseconds, spanning a millisecond to the {@link
    * RetentionPolicy#MAX_RUN} ceiling.
    *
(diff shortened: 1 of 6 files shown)


aaylward:
All three taken in e6c5ff5.

**1. Wrong issue.** Correct — #1313 is the first-load latency checklist, and this bug only landed near it because the branch name did. Filed **#1323** for the real incident (counters invisible for their first event, on every service, with the deployed evidence and the parts left open: other emitters unswept, distributions have the same shape, no encoder pin). `CustomMetrics`, the registry comment and the test now cite it.

**2. The volumes test re-coupled the constants** — the sharpest of the three, because the test contradicted the argument the same change made for splitting them. It now pins `burstWindow` itself against a **12-hour floor** (the reported gap is overnight-to-morning, so anything shorter differs from the default and still answers zero), and skips the alarm trio entirely, which `TestRegistry_AlarmCountersCountOverALongWindow` already owns.

Verified in both directions rather than by inspection:
- `alarmWindow: 24h → 6h` now **survives** the volumes test (the decoupling holds) while still being **killed** by the full suite (the alarm test still catches it).
- `burstWindow: 24h → 5m` is **killed**.

**3. Stale javadoc** — `IndexWorkerMetricsTest` never existed; pins are in `IndexWorkerTest`. Fixed.

**On the Trends note**: agreed and worth stating, so I've added it to the PR body — the `tsCounter` panels are unchanged, and a 24h Trends range already shows an overnight bump as a step-sized `increase`. The tile is what was blind, not the chart. The encoder pin for a declared-at-zero counter surviving OTLP as `asInt: "0"` is noted as an open item on #1323 rather than added here.

---

Issue #1323, opened 2026-08-07
Custom counters are invisible for their first event, on every service


aaylward:
## Symptom

`muchq.com/metrics/one_d4` showed an Indexing panel of zeros after an overnight run that indexed over a thousand games — not stale numbers, *zeros*, indistinguishable from a service that had never indexed anything.

## Cause

A `CustomMetrics` counter only exists once something increments it (`counters.computeIfAbsent` on the first `add`). So the series is **born already carrying** the value of the first event it counted. `increase()` and `rate()` measure the change *between* samples, and there is no earlier sample to measure from — so that first event shows no increase at all, and every sample after it is flat until something else happens.

The counter is not wrong: `games_indexed_total` really did read 1092. Every dashboard built on it is.

This is not a corner case. It is what every process does after every deploy, so **on every rail the first run of anything is uncounted** — it just took an emitter whose events are rare enough for "first run" and "only run so far" to coincide for anyone to notice.

## Evidence from the deployment

- The reported run completed at **2026-08-07T05:09:27Z** (derived from the request's `expiresAt`, which is retention-dated).
- `games_indexed_total{service_name="one_d4"}` first appears in Prometheus at **05:14** — the next scrape.
- Across **seven days** the metric has exactly one nonzero point: a run triggered by hand while investigating, visible only because the series already existed by then.
- The same service's `http_server_*` family is nonzero throughout, which is what made this look like a broken custom-metrics pipeline rather than a counter-lifecycle problem.

## Fixed in #1322

`CustomMetrics.defineCounter` declares a series so it exports as 0 from process start — the counter equivalent of the existing `defineDistribution` — and `IndexWorker` declares its bounded series at construction. That PR also widens the one_d4 indexing tiles from a 5m to a 24h window, which is a separate cause of the same report: 5m is the right window for serving traffic and the wrong one for work a person triggers.

## Left open deliberately

- **Other emitters have not been swept.** golf_hub, microgpt-serve and portrait all record custom counters and none of them declare a baseline, so each still loses its first event after every deploy. one_d4 was fixed because it was the reported one.
- **Distributions have the same shape.** `defineDistribution` only stores bounds; the `_count`/`_sum` series still spring into existence on first observation, so a windowed mean over the first run reads from a single sample. Less visible than the counter case and not addressed.
- **No encoder pin** that a declared-but-untouched counter survives OTLP as `asInt: "0"`. Low risk given the existing path, but nothing asserts it.

PR #1324, opened 2026-08-07
metrics: sweep the zero baseline across both rails, and cover distributions (#1323)

aaylward:
Closes the three items #1323 left open when the one_d4 fix (#1322) landed.

## 1. The C++ rail has the same bug

Not a variant of it — the same one. `futility/otel` creates instruments lazily on the first `RecordCounter`, so a counter's first exported sample carries that event's value and `increase()` has nothing earlier to measure it against.

The deployment shows the shape rather than my inferring it: golf_hub's `session_starts` has existed since 08-03 with **no increase in it across seven days**, and `chat_message_count` appeared mid-week already carrying its first events.

`MetricsRecorder::DeclareCounter` records a **zero**, which is what actually materializes the series — creating the instrument alone registers no attribute set, so the SDK still exports nothing until a measurement arrives.

- **portrait** declares its five series in the `TracerService` constructor.
- **golf_hub** declares its eleven unlabelled counters when the handler is built.

`stream_commands{command}` and `stream_events{event}` are deliberately left out: their labels come from a generated union's case names rather than a list either file could state, and they fire continuously in any live session — the case where a missing first event costs least. Declaring one series per possible value is how a metric becomes a cardinality problem, which is the same line the Java rail already draws around motif names.

## 2. Distributions had the counter's problem one level down

`defineDistribution` declares **bounds**, not a series, so the first observation still created the series — and a windowed `rate(sum)/rate(count)` over that first run divides a one-sample rate by a one-sample rate. So `avg_run_seconds_1h` was blind to the first run for exactly the same reason the counters were.

`defineDistributionSeries` takes the label set, which is why it can't fold into `defineDistribution`: bounds are per name, but a series is per name **and** labels, and declaring bounds for `index_run_duration_micros` says nothing about which outcomes it will be sliced by. `IndexWorker` declares its two, `RUN_DURATION` once per outcome.

## 3. The encoder pin that was missing

A declared-but-untouched counter ships as `asInt: "0"` — not dropped as uninteresting, and not written as a bare number the collector would reject. A declared distribution ships as an empty histogram whose buckets already carry the declared shape, so the first observation lands in the right bucket instead of reshaping the histogram mid-flight.

## Tests

Every "declared at zero" assertion is paired with a **presence** assertion. `CounterTotal` and the Java `counter()` helper both sum an empty match to `0`, so the zero reads alone would pass against code that declared nothing — the assertion would have been vacuous in exactly the way this whole issue is about.

**Five mutations killed**: `DeclareCounter`'s record, golf_hub's `DeclareMetrics()` call, and each of the two distribution declarations. The distribution mutation initially **survived** and the test was strengthened until it didn't.

One process note: a `--test_filter` run reported PASSED before the test existed, because the filter matched nothing. I re-ran with `--test_output=all` to confirm the golf_hub test actually executes rather than trusting the green.

## Verification

21/21 across yodel, futility/otel, one_d4 workers, portrait and golf_hub; `scripts/format-all` run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AHNHh5jFLF4t6oFxUR1Fsg)_

### DIFF
--- a/domains/games/apis/golf_hub/hub_e2e_test.cc
+++ b/domains/games/apis/golf_hub/hub_e2e_test.cc
@@ -1485,6 +1485,33 @@ TEST_F(GolfHubStreamFixture, ChatMetricsCountOutcomesWithoutIdentifiers) {
 // The unavailable side of the append counter: the store cannot be
 // reached, the sender is told so, and nothing counts as stored or
 // delivered.
+
+// The zero baseline (#1323): a handler must put its unlabelled counters on the
+// wire at 0 when it is built, before any session. Without it each series is
+// born carrying its first event's value and increase() shows nothing for that
+// event, ever — so the first chat message, the first refused admission, the
+// first reaped seat after a deploy are all uncounted.
+//
+// stream_commands and stream_events are deliberately absent: their labels come
+// from a generated union's case names rather than a list the handler could
+// state, and they fire continuously in any live session, which is the case a
+// missing first event costs least.
+TEST_F(GolfHubStreamFixture, BuildingAHandlerDeclaresItsUnlabelledCountersAtZero) {
+  auto capture = MakeCapturingMetricsRecorder();
+  auto instance = BuildSecondInstance(vault_, store_, chat_store_, capture);
+  ASSERT_NE(instance, nullptr);
+
+  for (const char* name :
+       {"chat_appends", "chat_failures", "chat_history_replays", "chat_rows_delivered",
+        "restored_seats_reaped", "stream_admissions_refused", "stream_disconnects",
+        "stream_rate_limited", "stream_rejections", "stream_seats_expired", "stream_sessions"}) {
+    EXPECT_EQ(capture->CounterTotal(name), 0) << name;
+  }
+  // Presence, not just zero reads: CounterTotal sums an empty match to 0, so
+  // every assertion above passes against a handler that declared nothing.
+  EXPECT_GE(capture->Entries().size(), 11U);
+}
+
 TEST_F(GolfHubStreamFixture, AnUnreachableStoreCountsTheAppendAsUnavailable) {
   auto capture = MakeCapturingMetricsRecorder();
   auto instance = BuildSecondInstance(
--- a/domains/games/apis/golf_hub/hub_handler.cc
+++ b/domains/games/apis/golf_hub/hub_handler.cc
@@ -171,7 +171,30 @@ HubHandler::HubHandler(std::shared_ptr<TicketVault> vault, std::shared_ptr<cards
         options.grace_period = grace_period;
         options.on_expired = [this](const std::string& id) { OnExpired(id); };
         return options;
-      }()) {}
+      }()) {
+  DeclareMetrics();
+}
+
+// Every unlabelled counter this handler can report, declared at construction so
+// each exports a zero before it counts anything. Without a baseline the series
+// is born carrying its first event's value, and increase() has nothing earlier
+// to measure it against — so the first chat message, the first refused
+// admission, the first reaped seat after a deploy are all invisible (#1323).
+//
+// The two labelled counters, stream_commands{command} and stream_events{event},
+// are deliberately absent: their labels come from a generated union's case
+// names rather than a list this file could state, and they fire continuously in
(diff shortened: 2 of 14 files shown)


aaylward:
The blocking finding was right, and so were both non-blocking ones. Fixed in a5fde08.

## The blocking one

Confirmed by reading every `Count(...)` site: six of the eleven "unlabelled" counters are only ever emitted with a label. The bare declaration baselined an orphan while the real series stayed unbaselined — the #1323 bug, dressed as a fix for it. `DeclaredCounterSeries()` now names all eighteen in full.

Your reading of the tile queries needs one correction that does not change the conclusion. `stream_sessions_total` with no matcher is a *selector*, so it matches every label set, and `sum(increase(...))` over it does include the labelled series. But the labelled series is still born carrying its first event, and the declared `{}` series is a flat zero that adds nothing — so the first session after a deploy is still lost, and `Sessions/resumed` (which does pin `resumed="true"`) never sees its first at all. The fix stands.

`stream_rejections` I looked at separately and it is **not** what I claimed. Every `Reject()` reason is a compile-time literal — the ~15 inline ones, the `reason =` assignments, and the engine statuses reached through `dealt.status().message()`, which are literals in `libs/cards/golf`. Nothing interpolates client data. It stays undeclared, but for the honest reason: ~30 strings across two libraries, and a hand-kept list would rot on the first new rejection. It is also the *opposite* of the cheap case I filed it under — a well-behaved session emits none, so the first rejection after a deploy is exactly the event someone goes looking for, and `Activity/rejections` computes `increase()` over it. The comment now says that, and names the real fix (a bounded `kind` alongside the free text) as out of scope here.

## Where the first version was weaker than you said

You flagged the `Entries().size() >= 11` presence check. Replacing it exposed something worse, which the review panel caught: reading the roster from `DeclaredCounterSeries()` meant deleting an entry deleted its own assertion. Fourteen of eighteen series could be removed with the suite green.

Two changes:

- **`ExpectOnlyDeclaredCounterSeries` moved into the fixture's `TearDown`**, after the session closes. Every test in the suite now checks its own paths against the declaration — the bad-ticket admission, the rate limiter, the chat-store failures, and the disconnects the closes produce. One scripted session covered four of eighteen.
- **The zero-baseline test keeps its own literal copy of the roster**, so a deletion costs two edits instead of silently deleting its assertion. That copy is also the only guard on the reverse mistake — a series declared and never emitted, which is a permanently flat dashboard line and which nothing else can see. Stated as a limit in the comment rather than claimed as coverage.

Mutation-checked, including the three that slipped past the first version:

| Mutation | Before | After |
|---|---|---|
| Declare a labelled series bare | killed | killed |
| Delete `stream_admissions_refused` from the roster | **survived** | killed |
| Delete `chat_appends{result="unavailable"}` | **survived** | killed |
| Change an emit-site label spelling (`clean` → `graceful`) | **survived** | killed |

## Non-blocking 1 — `IndexWorkerTest`

Right. `defineDistributionSeries` loops all four `RUN_OUTCOMES`; the test pinned `completed`. All four now, with presence asserted separately from the count (`distributionCount` sums an empty stream to 0, so `isZero()` alone was vacuous). Mutation: pinning the production loop to `"completed"` is killed.

## Non-blocking 2 — microgpt-serve

Right that it was named in #1323 and dropped without a rationale. Swept: `AppMetrics::declare()` puts the three counters on the wire at 0 across both endpoints, with `AppMetrics::new()` delegating to a `with_meter` seam so a test can read an isolated provider back.

The two histograms are deliberately **not** declared, and this is a real difference between the rails rather than an oversight: `record(0.0)` is an observation, not a declaration. It leaves `_sum` alone and adds one to `_count`, so `avg_duration_ms` — `rate(_sum)/rate(_count)` — would read low for the window containing process start. The Java rail can baseline a distribution because `defineDistributionSeries` does `computeIfAbsent` on an empty count-0 `Distribution`; the OTel `Histogram` API has no equivalent. golf_hub's `chat_catch_up_rows` has the identical gap and now says so in the same terms.

The Rust test also got the same treatment as the C++ one — it swept three instruments by name, which could not have caught a *new* undeclared counter. It now enumerates every u64 sum in the export and compares the whole set.

## Non-blocking 3 — presence via entry count

Gone, replaced by named-series presence per the above.

---

